PR #16, opened 2020-03-10
Create CONTRIBUTING.md

CaitlinV39:
## Description
I wanted to add a contributing file. I took the one that FHIR Server had but wonder if we need to update. Can you review and provide comments on where things should change? I'm specifically thinking under Coding Style.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,31 @@
+# Contributing to FHIR Converter
+
+This document describes guidelines for contributing to the FHIR Converter repo.
+
+## Submitting Pull Requests
+
+- **DO** submit all changes via pull requests (PRs). They will be reviewed and potentially be merged by maintainers after a peer review that includes at least one of the team members.
+- **DO** give PRs short but descriptive names.
+- **DO** write a useful but brief description of what the PR is for.
+- **DO** refer to any relevant issues and use [keywords](https://help.github.com/articles/closing-issues-using-keywords/) that automatically close issues when the PR is merged.
+- **DO** ensure each commit successfully builds. The entire PR must pass all checks before it will be merged.
+- **DO** address PR feedback in additional commits instead of amending.
+- **DO** assume that [Squash and Merge](https://blog.github.com/2016-04-01-squash-your-commits/) will be used to merge the commits unless specifically requested otherwise.
+- **DO NOT** submit "work in progress" PRs. A PR should only be submitted when it is considered ready for review.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+- **DO** address the StyleCop.Analyzers errors.
+- **DO** follow the [.editorconfig](.editorconfig) and [stylecop.json](stylecop.json) settings.
+
+## Creating Issues
+
+- **DO** use a descriptive title that identifies the issue or the requested feature.
+- **DO** write a detailed description of the issue or the requested feature.
+- **DO** provide details for issues you create:
+  - Describe the expected and actual behavior.
+  - Provide any relevant exception message or OperationOutcome.
```
deepak-ms: "OperationOutcome" doesn't make sense here.

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+- **DO** give PRs short but descriptive names.
+- **DO** write a useful but brief description of what the PR is for.
+- **DO** refer to any relevant issues and use [keywords](https://help.github.com/articles/closing-issues-using-keywords/) that automatically close issues when the PR is merged.
+- **DO** ensure each commit successfully builds. The entire PR must pass all checks before it will be merged.
+- **DO** address PR feedback in additional commits instead of amending.
+- **DO** assume that [Squash and Merge](https://blog.github.com/2016-04-01-squash-your-commits/) will be used to merge the commits unless specifically requested otherwise.
+- **DO NOT** submit "work in progress" PRs. A PR should only be submitted when it is considered ready for review.
+- **DO NOT** mix independent and unrelated changes in one PR.
+
+## Coding Style
```
deepak-ms: This coding style info is for .net. I think we can remove this section for now unless @LTA-Thinking has any recommendation.

LTA-Thinking: Our coding style is enforced by the eslint rules we have. We could mention that contributors should run the linter before submitting a PR.

deepak-ms: linter is already run as part of checks (which is mentioned in PR submission guidelines), so we should be good with removing this section.

CONTRIBUTING.md
CaitlinV39: Updated!

CONTRIBUTING.md
CaitlinV39: Removed the entire coding style section

PR #22, opened 2020-03-30
Add math helpers

feordin:
## Description

Add math helper functions

## Related issues

Addresses [issue #71948].

## Testing

Unit tests for all math helper functions added.

### DIFF
--- a/src/lib/handlebars-converter/handlebars-helpers.js
+++ b/src/lib/handlebars-converter/handlebars-helpers.js
@@ -597,28 +597,28 @@ module.exports.external = [
     },
     {
         name: 'abs',
-        description: 'Returns the absolute value of a number.',
+        description: 'Returns the absolute value of a number: abs number',
         func: function (x) {
             return Math.abs(x);
         }
     },
     {
         name: 'ceil',
-        description: 'Returns the next largest whole number or integer.',
+        description: 'Returns the next largest whole number or integer: ceil number',
         func: function (x) {
             return Math.ceil(x);
         }
     },
     {
         name: 'floor',
-        description: 'Returns the largest integer less than or equal to a given number.',
+        description: 'Returns the largest integer less than or equal to a given number: floor number',
         func: function (x) {
             return Math.floor(x);
         }
     },
     {
         name: 'max',
-        description: 'Returns the largest of zero or more numbers.',
+        description: 'Returns the largest of zero or more numbers: max number1, number2, number3 . . .',
         func: function () {
             var args = [];
             for (var i = 0; i < arguments.length - 1; i++)
@@ -628,7 +628,7 @@ module.exports.external = [
     },
     {
         name: 'min',
-        description: 'Returns the lowest-valued number passed into it, or NaN if any parameter isn\'t a number and can\'t be converted into one.',
+        description: 'Returns the lowest-valued number passed into it, or NaN if any parameter isn\'t a number and can\'t be converted into one: min number1, number2, number3 . . .',
         func: function () {
             var args = [];
             for (var i = 0; i < arguments.length - 1; i++)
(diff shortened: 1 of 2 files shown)


src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -594,5 +594,81 @@ module.exports.external = [
         func: function (o) {
             return isNaN(o);
         }
+    },
+    {
+        name: 'abs',
+        description: 'Returns the absolute value of a number.',
```
deepak-ms: Let's use the existing description format (signature at end).

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -594,5 +594,81 @@ module.exports.external = [
...
+        }
+    },
+    {
+        name: 'min',
+        description: 'Returns the lowest-valued number passed into it, or NaN if any parameter isn\'t a number and can\'t be converted into one.',
+        func: function () {
+            var args = [];
+            for (var i = 0; i < arguments.length - 1; i++)
+                args[i] = arguments[i];
+            return Math.min(...args);
```
deepak-ms: nit : will arguments.slice(0, -1) work here?

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -594,5 +594,81 @@ module.exports.external = [
...
+        func: function (x) {
+            return Math.sign(x);
+        }
+    },
+    {
+        name: 'trunc',
+        description: 'Returns the integer part of a number by removing any fractional digits.',
+        func: function (x) {
+            return Math.trunc(x);
+        }
```
deepak-ms: Should we have add/subtract/multiply/divide too?

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -594,5 +594,81 @@ module.exports.external = [
         func: function (o) {
             return isNaN(o);
         }
+    },
+    {
+        name: 'abs',
+        description: 'Returns the absolute value of a number.',
+        func: function (x) {
+            return Math.abs(x);
```
deepak-ms: Should we throw a friendly error in case of exception?

src/lib/handlebars-converter/handlebars-helpers.js
feordin: Unfortunately, arguments object is not an array, it is object which behaves somewhat array like. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments 
So, it does not have a slice method.

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -594,5 +594,109 @@ module.exports.external = [
...
+    {
+        name: 'trunc',
+        description: 'Returns the integer part of a number by removing any fractional digits: trunc number',
+        func: function (x) {
+            return Math.trunc(x);
+        }
+    },
+    {
+        name: 'add',
+        description: 'add two numbers: + number1 number 2',
```
deepak-ms, line 676: will "+" work in template? or will "add" be required?

feordin: I did initially try +, - , /, and * for the math helpers, but ran into some issues.  We post process - for the PID-1 scenario.  Also the / is a special character and didn't parse well either.  I think + would work, but to be consistent with all of them, I left it as 'add'.  I am open to changing it.

src/lib/handlebars-converter/handlebars-helpers.js
feordin: Discussed with Deepak, and we'll leave the math functions returning NaN if unable to calculate a numeric value.

PR #30, opened 2020-06-09
Personal/yankun/handleDateFormat

yankunhuang-pku:
## Description

Fix the date format conversion 

## Related issues
Addresses #74191

## Testing

The reason for the incorrect date conversion is that the original code didn’t check the different input formats of date from CDA and just simply output all the date as UTC format by filling the zero. For example:

1.Just year + month (+day): 
- Input: `<effectiveTime value="199911"/>`
- Output: `"occurrenceDateTime": "1999-10-31T00:00:00.000Z "`

2.Bad input format 
- Input: `<effectiveTime value="19991"/>`
- Output: `"occurrenceDateTime": "1999-09-30T00:00:00.000Z "`

I added some validation for the inputs which will throw exceptions for bad inputs, fixed the bug according to the specs of FHIR (https://www.hl7.org/fhir/datatypes.html#dateTime). For example: 
1.Just year + month (+day):  
- Input: `<effectiveTime value="199911"/>`
- Output: `"occurrenceDateTime": "1999-11"`

2.Bad input format  (Now the converter will throw exceptions and inform the user)

3.UTC format
- Input: `<effectiveTime value="19991101020203200"/>`
- Output: `"occurrenceDateTime": "1999-11-01T02:02:03.200Z"`

### DIFF
--- a/src/lib/handlebars-converter/handlebars-helpers.js
+++ b/src/lib/handlebars-converter/handlebars-helpers.js
@@ -36,10 +36,11 @@ var getDate = function (dateTimeString) {
     var ds = dateTimeString.toString();
     if (!/^\d+$/.test(ds))
         return null;
-    if (0 <= ds.length <= 5 && ds.length == 7)
-        return ds;
     if (ds.length == 6)
         return ds.substring(0, 4) + '-' + ds.substring(4, 6);
+    if (ds.length <= 7)
+        return ds;
+    
     if (ds.length == 8)
         return ds.substring(0, 4) + '-' + ds.substring(4, 6) + '-' + ds.substring(6, 8);
 
--- a/src/lib/handlebars-converter/handlebars-helpers.js
+++ b/src/lib/handlebars-converter/handlebars-helpers.js
@@ -32,7 +32,15 @@ var normalizeSectionName = function (name) {
 };
 
 var getDate = function (dateTimeString) {
+    // handling the date format here
     var ds = dateTimeString.toString();
+    if (/^(\d{0,5}|\d{7})$/.test(ds))
+        return ds;
+    if (/^\d{6}$/.test(ds))
+        return ds.substring(0, 4) + '-' + ds.substring(4, 6);
+    if (/^\d{8}$/.test(ds))
+        return ds.substring(0, 4) + '-' + ds.substring(4, 6) + '-' + ds.substring(6, 8);
+
     ds = ds.padEnd(17, '0');
     var year = ds.slice(0, 4);
     var monthIndex = ds.slice(4, 6) - 1;
@@ -44,7 +52,7 @@ var getDate = function (dateTimeString) {
     var minute = ds.slice(10, 12);
     var second = ds.slice(12, 14);
     var millisecond = ds.slice(14, 17);
-    return (new Date(Date.UTC(year, monthIndex, day, hour, minute, second, millisecond)));
+    return (new Date(Date.UTC(year, monthIndex, day, hour, minute, second, millisecond))).toJSON();
 };
 
 module.exports.internal = {
@@ -660,13 +668,11 @@ module.exports.external = [
         func: function (date) {
             try {
                 var bd = date.toString();
-
-                // Should be 8 digits
-                if (!/^\d{8}$/.test(bd)) {
-                    return bd;
-                }
-
-                return bd.substring(0, 4) + '-' + bd.substring(4, 6) + '-' + bd.substring(6, 8);
+                if (/^\d{8,17}$/.test(bd)) 
+                    return bd.substring(0, 4) + '-' + bd.substring(4, 6) + '-' + bd.substring(6, 8);
+                if (/^\d{6}$/.test(bd))
+                    return bd.substring(0, 4) + '-' + bd.substring(4, 6);
+                return bd;
             }
             catch (err) {
                 return '';
@@ -685,7 +691,7 @@ module.exports.external = [
         description: 'Converts an YYYYMMDDHHmmssSSS string, e.g. 20040629175400000 to dateTime format, e.g. 2004-06-29T17:54:00.000z: formatAsDateTime(dateTimeString)',
         func: function (dateTimeString) {
             try {
-                return getDate(dateTimeString).toJSON();
+                return getDate(dateTimeString);
             }
             catch (err) {
                 return '';
--- a/src/lib/handlebars-converter/handlebars-helpers.js
+++ b/src/lib/handlebars-converter/handlebars-helpers.js
@@ -40,7 +40,6 @@ var getDate = function (dateTimeString) {
         return ds.substring(0, 4) + '-' + ds.substring(4, 6);
     if (ds.length <= 7)
         return ds;
-    
     if (ds.length == 8)
         return ds.substring(0, 4) + '-' + ds.substring(4, 6) + '-' + ds.substring(6, 8);
 
--- a/src/lib/handlebars-converter/handlebars-helpers.js
+++ b/src/lib/handlebars-converter/handlebars-helpers.js
@@ -34,11 +34,13 @@ var normalizeSectionName = function (name) {
 var getDate = function (dateTimeString) {
     // handling the date format here
     var ds = dateTimeString.toString();
-    if (/^(\d{0,5}|\d{7})$/.test(ds))
+    if (!/^\d+$/.test(ds))
+        return null;
+    if (0 <= ds.length <= 5 && ds.length == 7)
         return ds;
-    if (/^\d{6}$/.test(ds))
+    if (ds.length == 6)
         return ds.substring(0, 4) + '-' + ds.substring(4, 6);
-    if (/^\d{8}$/.test(ds))
+    if (ds.length == 8)
         return ds.substring(0, 4) + '-' + ds.substring(4, 6) + '-' + ds.substring(6, 8);
 
     ds = ds.padEnd(17, '0');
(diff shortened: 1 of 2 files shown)


src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -34,7 +34,15 @@ var normalizeSectionName = function (name) {
 }
 
 var getDate = function (dateTimeString) {
+    // handling the date format here
     var ds = dateTimeString.toString();
+    if (/^\d{0,5}$/.test(ds) || /^\d{7}$/.test(ds))
```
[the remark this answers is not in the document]
yankunhuang-pku: yes, I have changed it!

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -32,7 +32,15 @@ var normalizeSectionName = function (name) {
...
 var getDate = function (dateTimeString) {
+    // handling the date format here
     var ds = dateTimeString.toString();
+    if (/^(\d{0,5}|\d{7})$/.test(ds))
+        return ds;
+    if (/^\d{6}$/.test(ds))
+        return ds.substring(0, 4) + '-' + ds.substring(4, 6);
+    if (/^\d{8}$/.test(ds))
+        return ds.substring(0, 4) + '-' + ds.substring(4, 6) + '-' + ds.substring(6, 8);
+
```
deepak-ms, line 43: Is it possible to optimize the logic here? Can we first check for all digits and then check against length and/or move the common case to the top?

yankunhuang-pku: OK. I changed the logic by checking the digits and then checking against length, like:
```
    if (!/^\d+$/.test(ds))
        return null;
    if (0 <= ds.length <= 5 && ds.length == 7)
        return ds;
    if (ds.length == 6)
        return ds.substring(0, 4) + '-' + ds.substring(4, 6);
    if (ds.length == 8)
        return ds.substring(0, 4) + '-' + ds.substring(4, 6) + '-' + ds.substring(6, 8);
```

src/lib/handlebars-converter/handlebars-helpers.spec.js
```diff
@@ -104,8 +104,12 @@ describe('Handlebars helpers', function () {
         { f: 'addHyphensSSN', in: [undefined], out: "" },
         { f: 'addHyphensDate', in: [undefined], out: "" },
         { f: 'formatAsDateTime', in: [undefined], out: "" },
+        {f: 'formatAsDateTime', in: ["20041wrong"], out: null}, // eslint-disable-line
+        {f: 'formatAsDateTime', in: ["2004"], out: "2004"}, // eslint-disable-line
+        {f: 'formatAsDateTime', in: ["20041"], out: "20041"}, // eslint-disable-line
```
deepak-ms, line 109: "2004wrong" and "20041" (both incorrect) should be treated in same manner (preferably return empty). @ranvijaykumar

deepak-ms: we need to have the same behavior from addHyphensDate too: either return input string or empty.

LTA-Thinking: I think the user needs to be informed that there is a problem with the conversion. With our current framework I think an error is the best solution, but it might be worth thinking about a way to return a list of warnings so the user has the choice to continue or not.

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -660,13 +670,13 @@ module.exports.external = [
...
-                }
-
-                return bd.substring(0, 4) + '-' + bd.substring(4, 6) + '-' + bd.substring(6, 8);
+                if (!/^\d+$/.test(bd))
+                    return null;
+                if (bd.length == 6)
+                    return bd.substring(0, 4) + '-' + bd.substring(4, 6);
+                if (bd.length >= 8)
+                    return bd.substring(0, 4) + '-' + bd.substring(4, 6) + '-' + bd.substring(6, 8);
+                return bd;
```
LTA-Thinking, line 679: Since the date parsing logic is the same here and above could it be moved to a shared function? If not I think making the regex into a constant is still a good idea.

src/lib/handlebars-converter/handlebars-helpers.spec.js
yankunhuang-pku: Yes, I think so. I think it's necessary to validated the input and output data and prompted the user in a unified way for the errors in the conversion process.

src/lib/handlebars-converter/handlebars-helpers.js
yankunhuang-pku: Thanks, I can unify this part of processing logic for date and datetime, and then extract a shared function. My previous idea is to make different logic for date and datetime. For Date, if the length of digits exceeds 8, I hope to keep the first 8 digits, such as 1999010101, which will be converted to 1999-01-01, but for DateTime, it will be filled with zero to 17 digits.

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -31,26 +31,137 @@ var normalizeSectionName = function (name) {
...
+        if (ds.length <= 8)
+            return convertDate(ds);
+        
+        // Padding 0s to 17 digits 
+        dateTimeComposition = getDateTimeComposition(ds);
+        if (!validUTCDateTime(dateTimeComposition))
+            throw `Invalid datetime: ${ds}`;
+        return (new Date(Date.UTC(dateTimeComposition.year, dateTimeComposition.month - 1, dateTimeComposition.day, dateTimeComposition.hours, dateTimeComposition.minutes, dateTimeComposition.seconds, dateTimeComposition.milliseconds))).toJSON();
+    }
+    catch (err) {
```
LTA-Thinking, line 156: The try catch here isn't needed since it is just rethrowing the error.

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -31,26 +31,137 @@ var normalizeSectionName = function (name) {
...
+
+// handling the date format here
+var getDate = function (dateString) {
+    try{
+        if (!validString(dateString))
+            return '';
+        var ds = validDigitString(dateString);
+        return convertDate(ds);
+    }
+    catch (err) {
```
LTA-Thinking, line 96: The try catch here isn't needed since it is just rethrowing the error.

src/lib/handlebars-converter/handlebars-helpers.js
yankunhuang-pku: Thanks, I have removed it.

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -31,26 +31,137 @@ var normalizeSectionName = function (name) {
     return name.replace(/[^A-Za-z0-9]/g, '_');
 };
 
-var getDate = function (dateTimeString) {
+// check the date is valid
+var validDate = function (year, monthIndex, day){
+    var dateInstance = new Date(year, monthIndex, day);
+    if(dateInstance.getFullYear() == year && dateInstance.getMonth() == monthIndex && dateInstance.getDate() == day)
```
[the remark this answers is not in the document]
yankunhuang-pku: Thanks, I have changed it.

src/lib/handlebars-converter/handlebars-helpers.spec.js
```diff
@@ -253,14 +280,22 @@ describe('Handlebars helpers', function () {
...
+        assert.strictEqual('2001-06', getHelper('addHyphensDate').func('200106'));
+        assert.strictEqual('2001-12', getHelper('addHyphensDate').func('200112'));
+    });
+
+    it('addHyphensDate adds hyphens when passed more than 8 digits', function() {
+        assert.strictEqual('2001-12-01', getHelper('addHyphensDate').func('20011201'));
+        assert.strictEqual('2001-12-31', getHelper('addHyphensDate').func('20011231'));
+        assert.strictEqual('2001-01-31', getHelper('addHyphensDate').func('20010131'));
+        assert.strictEqual('2001-01-31', getHelper('addHyphensDate').func('2001013100'));
+        assert.strictEqual('2001-01-31', getHelper('addHyphensDate').func('200101310000'));
```
deepak-ms, line 298: Can we move these 3 tests to above group test (along with formatAsDateTime)?

yankunhuang-pku: Yes, I also think they should be put on group test. I just follow the previous position. If it doesn't affect others, I move them to group test

src/lib/handlebars-converter/handlebars-helpers.spec.js
```diff
@@ -222,7 +258,7 @@ describe('Handlebars helpers', function () {
         var currentString = getHelper('now').func();
         var after = new Date();
 
-        var current = helperUtils.getDate(currentString);
+        var current = new Date(helperUtils.getDateTime(currentString));
```
deepak-ms, line 261: why is this change needed?

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -31,26 +31,129 @@ var normalizeSectionName = function (name) {
...
+    dateTimeComposition = getDateTimeComposition(ds);
+    if (!validUTCDateTime(dateTimeComposition))
+        throw `Invalid datetime: ${ds}`;
+    return (new Date(Date.UTC(dateTimeComposition.year, dateTimeComposition.month - 1, dateTimeComposition.day, dateTimeComposition.hours, dateTimeComposition.minutes, dateTimeComposition.seconds, dateTimeComposition.milliseconds))).toJSON();
+};
+
+module.exports.internal = {
+    getDateTime: getDateTime
+};
+
```
deepak-ms, line 156: Does it override the declaration above for getDate?

src/lib/handlebars-converter/handlebars-helpers.spec.js
yankunhuang-pku: because the type of data the getDateTime return is a string, I use toJSON() in return.

src/lib/handlebars-converter/handlebars-helpers.js
yankunhuang-pku: Yes, I modified it

PR #38, opened 2020-06-24
Update the bad input handling

yankunhuang-pku:
## Description

Support the datetime with the time zone information (Format: YYYY[MM[DD[HH[MM[SS[.S[S[S[S]]]]]]]]][+/-ZZZZ].)

Update the bad input handling according to the spec in http://www.hl7.eu/refactored/dtDTM.html

## Testing

Add some test cases

### DIFF
--- a/src/lib/handlebars-converter/handlebars-helpers.js
+++ b/src/lib/handlebars-converter/handlebars-helpers.js
@@ -56,15 +56,11 @@ var validUTCDateTime = function (dateTimeComposition){
 var validString = function (dateTimeString) {
     if (!dateTimeString || dateTimeString.toString() === '')
         return false;
-    return true;
-};
-
-// check the characters of string are digits
-var validDigitString = function (dateTimeString) {
+    // datetime format in the spec: YYYY[MM[DD[HH[MM[SS[.S[S[S[S]]]]]]]]][+/-ZZZZ],
     var ds = dateTimeString.toString();
-    if (!/^(\d+)$/.test(ds))
-        throw `Contains non-numeric characters in ${ds}`;
-    return ds;
+    if (!/^((\d{4}(\d{2}(\d{2}(\d{2}(\d{2}(\d{2}(\.\d+)?)?)?)?)?)?((-|\+)\d{1,4})?))$/.test(ds))
+        throw `Bad input for Datetime type in ${ds}`;
+    return true;
 };
 
 // convert the dateString to date string with hyphens
@@ -91,11 +87,11 @@ var convertDate = function (dateString) {
 var getDate = function (dateString) {
     if (!validString(dateString))
         return '';
-    var ds = validDigitString(dateString);
-    return convertDate(ds);
+    return convertDate(dateString.toString());
 };
 
 var getDateTimeComposition = function (ds){
+    ds = ds.replace('.','');
     ds = ds.padEnd(17, '0');
     var year = ds.substring(0,4);
     var month = ds.substring(4,6);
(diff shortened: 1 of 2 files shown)


src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -56,15 +56,11 @@ var validUTCDateTime = function (dateTimeComposition){
...
-};
-
-// check the characters of string are digits
-var validDigitString = function (dateTimeString) {
+    // datetime format in the spec: YYYY[MM[DD[HH[MM[SS[.S[S[S[S]]]]]]]]][+/-ZZZZ],
     var ds = dateTimeString.toString();
-    if (!/^(\d+)$/.test(ds))
-        throw `Contains non-numeric characters in ${ds}`;
-    return ds;
+    if (!/^((\d{4}(\d{2}(\d{2}(\d{2}(\d{2}(\d{2}(\.\d+)?)?)?)?)?)?((-|\+)\d{1,4})?))$/.test(ds))
```
LTA-Thinking, line 61: Nit: You have an extra set of parenthesis for the group between the ^ and $.

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -56,15 +56,11 @@ var validUTCDateTime = function (dateTimeComposition){
 var validString = function (dateTimeString) {
```
LTA-Thinking, line 56: Could you change the name to something like 'validDateString'? This method is now testing specifically for date formats so the name should reflect that.

src/lib/handlebars-converter/handlebars-helpers.js
yankunhuang-pku: Thanks~

src/lib/handlebars-converter/handlebars-helpers.js
yankunhuang-pku: Thanks~ It's a good suggestion after my changes. I have renamed the function name.

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -53,18 +53,14 @@ var validUTCDateTime = function (dateTimeComposition){
...
-};
-
-// check the characters of string are digits
-var validDigitString = function (dateTimeString) {
+    // datetime format in the spec: YYYY[MM[DD[HH[MM[SS[.S[S[S[S]]]]]]]]][+/-ZZZZ],
     var ds = dateTimeString.toString();
-    if (!/^(\d+)$/.test(ds))
-        throw `Contains non-numeric characters in ${ds}`;
-    return ds;
+    if (!/^(\d{4}(\d{2}(\d{2}(\d{2}(\d{2}(\d{2}(\.\d+)?)?)?)?)?)?((-|\+)\d{1,4})?)$/.test(ds))
```
LTA-Thinking, line 61: Looking at the spec the time zone has to be 4 digits. I think the time zone check should be d{4} instead of d{1,4}.

yankunhuang-pku: Ranvijay has noticed that another spec for the ccda is written as follows. Here I want to support 1~4 digits for timezone, because it may exist, and I don't want to throw exceptions too strictly.
```
<xs:simpleType name="ts">
<xs:annotation>
<xs:documentation> A quantity specifying a point on the axis of natural time. A point in time is most often represented as a calendar expression. </xs:documentation>
</xs:annotation>
<xs:restriction base="xs:string">
<xs:pattern value="[0-9]{1,8}|([0-9]{9,14}|[0-9]{14,14}\.[0-9]+)([+\-][0-9]{1,4})?"/>
</xs:restriction>
</xs:simpleType>
```

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -53,18 +53,14 @@ var validUTCDateTime = function (dateTimeComposition){
...
-var validString = function (dateTimeString) {
+var validDatetimeString = function (dateTimeString) {
     if (!dateTimeString || dateTimeString.toString() === '')
         return false;
-    return true;
-};
-
-// check the characters of string are digits
-var validDigitString = function (dateTimeString) {
+    // datetime format in the spec: YYYY[MM[DD[HH[MM[SS[.S[S[S[S]]]]]]]]][+/-ZZZZ],
```
moria97, line 59: Is there at most four digits in milliseconds? It seems not consistent with the regex below.

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -118,24 +114,28 @@ var getDateTimeComposition = function (ds){
 
 // handling the datetime format here
 var getDateTime = function (dateTimeString) {
-    if (!validString(dateTimeString))
+    if (!validDatetimeString(dateTimeString))
         return '';
 
-    // handle special datetime format like 20130130080051+0500, 20130130080051-0500
+    // handle the datetime format with time zone
     var ds = dateTimeString.toString();
```
moria97, line 121: Not sure the purpose of toString here?

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -118,24 +114,28 @@ var getDateTimeComposition = function (ds){
...
+    var timeZoneChar = '';
+    if (ds.indexOf('-') !== -1)
+        timeZoneChar = '-';
+    else if (ds.indexOf('+') !== -1)
+        timeZoneChar = '+'; 
+    if (timeZoneChar !== '')
+    {
+        var dateSections = ds.split(timeZoneChar);
         var dateTimeComposition = getDateTimeComposition(dateSections[0]);
         var date = dateTimeComposition.year + '-' + dateTimeComposition.month + '-' + dateTimeComposition.day;
```
moria97, line 131: What if the dateTime is in "YYYY" format? Not sure whether the month and day values are valid.

PR #40, opened 2020-06-26
Deploy specific image

LTA-Thinking:
## Description

Changes the deployment template to target a specific image instead of latest.

## Related issues

Addresses Bug 74582: UI of old installations getting updated automatically.

## Testing

Ran deployment and confirmed version that was deployed.

### DIFF
--- a/deploy/default-azuredeploy.json
+++ b/deploy/default-azuredeploy.json
@@ -34,7 +34,7 @@
     },
     "imageTag": {
       "type": "string",
-      "defaultValue": "2020-06-12"
+      "defaultValue": "2020-06-08"
     }
   },
   "variables": {
--- a/deploy/default-azuredeploy.json
+++ b/deploy/default-azuredeploy.json
@@ -35,10 +35,10 @@
     "imageTag": {
       "type": "string",
       "allowedValues": [
-        "1.0.0",
-        "2.0.0"
+        "v1.0.0",
+        "v2.0.0"
       ],
-      "defaultValue": "2.0.0"
+      "defaultValue": "v2.0.0"
     }
   },
   "variables": {
--- a/deploy/default-azuredeploy.json
+++ b/deploy/default-azuredeploy.json
@@ -35,10 +35,10 @@
     "imageTag": {
       "type": "string",
       "allowedValues": [
-        "2020-06-08",
-        "2020-06-29"
+        "1.0.0",
+        "2.0.0"
       ],
-      "defaultValue": "2020-06-08"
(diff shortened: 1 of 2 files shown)


deploy/default-azuredeploy.json
```diff
@@ -1,125 +1,129 @@
...
-                "clientAffinityEnabled": false,
-                "httpsOnly": true
+            {
+              "name": "WEBSITES_ENABLE_APP_SERVICE_STORAGE",
+              "value": "true"
             }
-        }
-    ]
+          ],
+          "linuxFxVersion": "[concat('DOCKER|', variables('registryName'), '/fhirconverter:', parameters('imageTag'))]",
```
moria97, line 119: Do we need to list some recommended image tags in the readme doc?

LTA-Thinking: That is a good idea. I think also making some images with clear names would be a good idea. We can limit the valid input for this field, which will give the user a dropdown list of options instead of a free text box.

moria97: Yes, we can modify this 'imageTag' parameter with some allowedValues.

LTA-Thinking: I added two versions for the moment: 2020-06-08 which was the last v1 image, and 2020-06-29 which is the most recent image at this time.

deploy/default-azuredeploy.json
```diff
@@ -1,125 +1,133 @@
...
+        "P1",
+        "P2",
+        "P3"
+      ],
+      "defaultValue": "B2"
+    },
+    "imageTag": {
+      "type": "string",
+      "allowedValues": [
+        "2020-06-08",
```
moria97, line 38: Here we can use v1.0.0 & v2.0.0

deploy/default-azuredeploy.json
```diff
@@ -1,125 +1,133 @@
...
+      ],
+      "defaultValue": "B2"
+    },
+    "imageTag": {
+      "type": "string",
+      "allowedValues": [
+        "1.0.0",
+        "2.0.0"
+      ],
+      "defaultValue": "2.0.0"
```
moria97, line 41: the versions should be v1.0.0 and v2.0.0

LTA-Thinking: Can't do the facepalm emoji here, right when I need it.

moria97: Never mind. I have tested that both options work perfectly now

PR #43, opened 2020-07-03
Personal/yufei/change docker image tag

moria97:
## Description

Add docker image tag with version info.

## Related issues

Addresses compatibility issue between different versions.

## Testing

Manually test CI pipeline in personal branch.


.github/workflows/dockerimage.yml
```diff
@@ -22,11 +22,15 @@ jobs:
         login-server: healthplatformregistry.azurecr.io
         username: ${{ secrets.REGISTRY_USERNAME }}
         password: ${{ secrets.REGISTRY_PASSWORD }}
-    
+
+    - name: get-npm-version
+      id: package-version
+      uses: martinbeentjes/npm-get-version-action@master
```
LTA-Thinking, line 28: Instead of pulling this from the master branch, should we pull the specific `@v1.0.0` version? It won't make a difference right now, but might avoid issues in the future.

moria97: Yes. As this pipeline runs only when master is updated,which means we are updating the newest code of the current version. We might need add branch trigger for old versions like v1.0.0/v2.0.0 respectively so that old images can get updated.

LTA-Thinking: I'm sorry, my comment wasn't specific. I meant pulling the 1.0.0 version of the npm-get-version-action instead of pulling its master version.
I agree with your statement though. Once we make the version specific branches we'll need to add them to this script.

moria97: Got it. I agree, using a specific version totally make sense. Thanks!

.github/workflows/dockerimage.yml
```diff
@@ -22,11 +22,15 @@ jobs:
...
+
     - name: Build the Docker image
       env:
         IMAGE_NAME: healthplatformregistry.azurecr.io/fhirconverter
       run: |
-        docker build . -t $IMAGE_NAME:${{ github.sha }} -t $IMAGE_NAME:$(date +'%Y-%m-%d') -t $IMAGE_NAME:latest
+        docker build . -t $IMAGE_NAME:${{ github.sha }} -t $IMAGE_NAME:v${{ steps.package-version.outputs.current-version }}-$(date +'%Y-%m-%d') -t $IMAGE_NAME:latest
```
deepak-ms, line 34: How about removing date from tag here (keeping only vmajor.minor.patch format), and adding the version to available list only when it's stable (after switching to next minor/patch version)?

moria97: Agree. This version syntax is more semantic and powerful. In this way, we should increment the version number in package.json during each update, right?

deepak-ms: Only thing we need to take care of is to increment version (major or minor or patch) after adding previous one in available list. Increment version after each update (internal version) doesn't impact customers, so it's not a strict requirement.

moria97: Sounds good. Thanks!

PR #44, opened 2020-07-06
Personal/yankun/e2e test framework

yankunhuang-pku:
## Description

Add the e2e test framework by validating the output with some rules.

In this framework, 
1. A rule lib is created in src/e2e-test/test-rule-functions.js. Now, I just give some samples and we can continue to add more rules in the future.
2.  We can customize some rule sets in src/e2e-test/config/rule-sets.js.
3.  We can customize some test cases in  src/e2e-test/config/testcases-cda.js or src/e2e-test/config/testcases-hl7v2.js

## Testing

Add some functional and unit test.

### DIFF
--- a/src/e2e-test/config/testcases-hl7v2.js
+++ b/src/e2e-test/config/testcases-hl7v2.js
@@ -1,0 +1,17 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+var ruleSets = require('./rule-sets'); 
+var hl7v2Testcases = [
+    { dataFile: 'ADT01-28.hl7', templateFile: 'ORU_R01.hbs', testRules: ruleSets.commonRules},
+
+    // error templeates with default GUID
+    { dataFile: 'ADT01-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
+    { dataFile: 'ADT04-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
+    { dataFile: 'ADT04-251.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
+];
+
+module.exports = hl7v2Testcases;
+
--- a/src/e2e-test/config/testcases-hl7v2.js
+++ b/src/e2e-test/config/testcases-hl7v2.js
@@ -8,9 +8,9 @@ var hl7v2Testcases = [
     { dataFile: 'ADT01-28.hl7', templateFile: 'ORU_R01.hbs', testRules: ruleSets.commonRules},
 
     // error templeates with default GUID
-    { dataFile: 'ADT01-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
-    { dataFile: 'ADT04-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
-    { dataFile: 'ADT04-251.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
+    // { dataFile: 'ADT01-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
+    // { dataFile: 'ADT04-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
+    // { dataFile: 'ADT04-251.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
 ];
 
 module.exports = hl7v2Testcases;
--- a/src/e2e-test/config/testcases-hl7v2.js
+++ b/src/e2e-test/config/testcases-hl7v2.js
@@ -6,12 +6,7 @@
 // design the test cases
 var ruleSets = require('./rule-sets'); 
 var hl7v2Testcases = [
-    { dataFile: 'ADT01-28.hl7', templateFile: 'ORU_R01.hbs', testRules: ruleSets.commonRules},
-
-    // error templeates with default GUID
-    // { dataFile: 'ADT01-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.GUIDRules},
-    // { dataFile: 'ADT04-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.GUIDRules},
-    // { dataFile: 'ADT04-251.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.GUIDRules},
+    { dataFile: 'ADT01-28.hl7', templateFile: 'ORU_R01.hbs', testRules: ruleSets.commonRules}
 ];
 
 module.exports = hl7v2Testcases;
--- a/src/e2e-test/config/testcases-hl7v2.js
+++ b/src/e2e-test/config/testcases-hl7v2.js
@@ -3,14 +3,15 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+// design the test cases
 var ruleSets = require('./rule-sets'); 
 var hl7v2Testcases = [
     { dataFile: 'ADT01-28.hl7', templateFile: 'ORU_R01.hbs', testRules: ruleSets.commonRules},
 
     // error templeates with default GUID
-    // { dataFile: 'ADT01-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
-    // { dataFile: 'ADT04-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
-    // { dataFile: 'ADT04-251.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.commonRules},
+    // { dataFile: 'ADT01-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.GUIDRules},
+    // { dataFile: 'ADT04-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.GUIDRules},
+    // { dataFile: 'ADT04-251.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.GUIDRules},
 ];
 
 module.exports = hl7v2Testcases;
--- a/src/e2e-test/e2e-test.spec.js
+++ b/src/e2e-test/e2e-test.spec.js
@@ -1,0 +1,73 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+const supertest = require('supertest');
+const express = require('express');
+var fs = require("fs");
+var path  = require('path');
+var constants = require('../lib/constants/constants');
+const API_KEY_HEADER = "X-MS-CONVERSION-API-KEY";
+const apiKeys = ['123', '456'];
+var app = require("../routes")(express());
+
+// load testcases
+var cdaTestcases = require('./config/testcases-cda');
+var hl7v2Testcases = require('./config/testcases-hl7v2');
+var opTests = cdaTestcases.concat(hl7v2Testcases);
+
+describe('E2E test - FHIR data validation', function () {
+    before(function () {
+        app.setValidApiKeys(apiKeys);
+    });
+
+    opTests.forEach(t => {
+        it('should output a valid FHIR data with ' + t.dataFile + ' and ' + t.templateFile, function (done) {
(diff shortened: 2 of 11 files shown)


src/e2e-test/e2e-test.spec.js
```diff
@@ -0,0 +1,73 @@
...
+            if(dataType === '.hl7'){
+                endpointURL = '/api/convert/hl7v2';
+                templateLocation = constants.HL7V2_TEMPLATE_LOCATION;
+                dataLocation = constants.HL7V2_DATA_LOCATION;
+            }
+            else if(dataType === '.cda'){
+                endpointURL = '/api/convert/cda';
+                templateLocation = constants.CDA_TEMPLATE_LOCATION;
+                dataLocation = constants.CDA_DATA_LOCATION;
+            }
```
moria97, line 45: Handle case for else?

src/e2e-test/config/testcases-hl7v2.js
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+// design the test cases
+var ruleSets = require('./rule-sets'); 
+var hl7v2Testcases = [
+    { dataFile: 'ADT01-28.hl7', templateFile: 'ORU_R01.hbs', testRules: ruleSets.commonRules},
+
+    // error templeates with default GUID
+    // { dataFile: 'ADT01-23.hl7', templateFile: 'ADT_A01.hbs', testRules: ruleSets.GUIDRules},
```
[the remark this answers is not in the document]
yankunhuang-pku: I just want to indicate that there is something wrong with this template. I've removed it.

src/e2e-test/testUtils.js
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+const countOccurences = (arr, value) => arr.reduce((a, v) => v === value ? a + 1 : a + 0, 0);
+const defaultGUID = '4cfe8d6d-3fc8-3e41-b921-f204be18db31'; // pass the parameter 'undefined'
+
+var findDuplicates = function(arr) {
+    var result = [];
+    arr.forEach(function(elem){
+        if(arr.indexOf(elem) !=arr.lastIndexOf(elem) && result.indexOf(elem) == -1){
```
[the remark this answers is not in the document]
yankunhuang-pku: I have optimize this function.

src/e2e-test/test-rule-functions.js
```diff
@@ -0,0 +1,50 @@
...
+    if (!result.valid)
+        return JSON.stringify(result, null, "\t");
+    return '';
+};
+
+var onePatient = function(resJson){
+    var resources = fhir.evaluate(resJson, 'Bundle.entry.resource.resourceType');
+    var patientCount = testUtils.countOccurences(resources,'Patient');
+    if( patientCount !== 1)
+        return 'The bundle contains ' + patientCount + ' Patient resources';
```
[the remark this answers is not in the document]
yankunhuang-pku: Yes, it will be more extensible. I have modified them.

src/e2e-test/e2e-test.spec.js
yankunhuang-pku: Thanks~I have modified it and it will throw an error when the data type is not supported.

src/e2e-test/test-rule-functions.js
```diff
@@ -0,0 +1,50 @@
...
+    var defaultGUIDCount = testUtils.countOccurences(ids, testUtils.defaultGUID);
+    if(defaultGUIDCount > 0)
+        return 'The bundle contains ' + defaultGUIDCount + ' default GUID ' + testUtils.defaultGUID;
+    else
+        return '';
+};
+
+var noSameGUID = function(resJson){
+    var ids = fhir.evaluate(resJson, 'Bundle.entry.resource.id');
+    var duplicates = testUtils.findDuplicates(ids);
```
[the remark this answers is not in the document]
yankunhuang-pku: Based on the current implementation of helper generateUUID, I don't think it's likely to generate the same ID.

src/e2e-test/e2e-test.spec.js
```diff
@@ -0,0 +1,76 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+const supertest = require('supertest');
+const express = require('express');
+var fs = require("fs");
+var path  = require('path');
+var constants = require('../lib/constants/constants');
+const API_KEY_HEADER = "X-MS-CONVERSION-API-KEY";
+const apiKeys = ['123', '456'];
```
LTA-Thinking, line 12: Why do we need two api keys?

src/e2e-test/test-rule-functions.spec.js
```diff
@@ -0,0 +1,268 @@
...
+                    }
+                }
+            ]
+        };
+        assert.strictEqual(testRules.fhirR4Validation(resJson).valid, false);
+        assert.strictEqual(JSON.parse(testRules.fhirR4Validation(resJson).errorMessage).valid, false);
+    });
+
+    it('Rule onePatient should return empty string when there is one Patient resourse', function () {
+        var resJson = {
```
LTA-Thinking, line 54: Since this bundle is used for several tests could it be made a global variable to save repetition?

src/e2e-test/test-rule-functions.js
```diff
@@ -0,0 +1,50 @@
...
+        return { valid: false, errorMessage: 'The bundle contains ' + defaultGuidCount + ' default Guid ' + testUtils.defaultGuid};
+    else
+        return { valid: true, errorMessage: ''};
+};
+
+var noSameGuid = function(resJson){
+    var ids = fhir.evaluate(resJson, 'Bundle.entry.resource.id');
+    var duplicates = testUtils.findDuplicates(ids);
+    if(duplicates.length !== 0)
+        return { valid: false, errorMessage: 'The bundle contains some duplicate Guid: ' + duplicates.toString()};
```
LTA-Thinking, line 39: Nit: Guid should be plural.

LTA-Thinking: Same on line 30

src/e2e-test/test-rule-functions.js
```diff
@@ -0,0 +1,50 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+var testUtils = require('./testUtils');
+var fhirModule = require('fhir');
+var fhir = new fhirModule.Fhir();
+
+var fhirR4Validation =  function(resJson){
```
LTA-Thinking, line 10: We have this rule, but neither of the current rule sets use it. Should it be used in a test? It seems valuable for template verification.

src/e2e-test/config/testcases-hl7v2.js
```diff
@@ -0,0 +1,13 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+// design the test cases
+var ruleSets = require('./rule-sets'); 
+var hl7v2Testcases = [
+    { dataFile: 'ADT01-28.hl7', templateFile: 'ORU_R01.hbs', testRules: ruleSets.commonRules}
```
LTA-Thinking, line 9: I feel like every top level template should have at least one testcase for it. As this PR is setting up the framework I see adding more tests as a separate work item, but I think that work item is needed.

src/e2e-test/e2e-test.spec.js
yankunhuang-pku: Just for example, we can support multiple keys

src/e2e-test/test-rule-functions.js
yankunhuang-pku: Yeah. I didn't use it in all rule sets because all the current conversion results can't pass the test now. However, I have some unit-tests in test-rule-functions.spec.js.

src/e2e-test/config/testcases-hl7v2.js
yankunhuang-pku: Yes, I agree with you. However, because some templates have errors now like 'ADT_A01.hbs', npm test will fail if it is added at present.

src/e2e-test/test-rule-functions.spec.js
yankunhuang-pku: Right, thanks for your reminder. I have modified it.

src/e2e-test/test-rule-functions.js
yankunhuang-pku: I have modified it. Thanks

src/e2e-test/e2e-test.spec.js
```diff
@@ -0,0 +1,76 @@
...
+
+    opTests.forEach(t => {
+        it('should output a valid FHIR data with ' + t.dataFile + ' and ' + t.templateFile, function (done) {
+            var dataType = path.extname(t.dataFile);
+            var endpointURL = '';
+            var templateLocation = '';
+            var dataLocation = '';
+            var requestJson = {
+                templateBase64: "",
+                messageBase64: ""
```
deepak-ms, line 33: there is no messageBase64 field now.

src/e2e-test/test-rule-functions.js
deepak-ms: It might be tough to reach zero validation error (to enable this test) and stay there consistently. Is it possible to add upper limit on validation failures for this? So that we can at least start using this test with some limit and ensure subsequent changes don't increase the validation failure count.

src/e2e-test/test-rule-functions.js
```diff
@@ -0,0 +1,56 @@
...
+        return response(true);
+};
+
+var noDefaultGuid = function(resJson){
+    var ids = fhir.evaluate(resJson, 'Bundle.entry.resource.id');
+    var defaultGuidCount = testUtils.countOccurences(ids, testUtils.defaultGuid);
+    if(defaultGuidCount === 1)
+        return response(false, 'The bundle contains ' + defaultGuidCount + ' default Guid ' + testUtils.defaultGuid);
+    else if(defaultGuidCount > 1)
+        return response(false, 'The bundle contains ' + defaultGuidCount + ' default Guids ' + testUtils.defaultGuid);
```
deepak-ms, line 36: why do we need to differentiate between 1 and > 1?

src/e2e-test/test-rule-functions.js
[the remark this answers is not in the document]
deepak-ms: As per fhir spec, resource id can be same across diff resource types. We should check for duplicates within a type.

src/e2e-test/test-rule-functions.js
yankunhuang-pku: We can use this rule by modifying templates slightly instead of adding upper limits because most of them just lack the necessary properties, such as test case `dataFile: 'C-CDA_R2-1_CCD.xml.cda', templateFile: 'ccd.hbs'`, only these two errors need to be corrected:
```
{
        "location": "Coverage.beneficiary",
        "resourceId": "ebd801b5-9113-3479-b35b-71002aae6c28",
        "severity": "error",
        "message": "Missing property"
},
{
        "location": "Coverage.payor",
        "resourceId": "ebd801b5-9113-3479-b35b-71002aae6c28",
        "severity": "error",
        "message": "Missing property"
}
```

src/e2e-test/e2e-test.spec.js
yankunhuang-pku: OK, I modified it.

src/e2e-test/test-rule-functions.js
yankunhuang-pku: As Robert said, Guid should be plural when >1.  (Guids)

src/e2e-test/test-rule-functions.js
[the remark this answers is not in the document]
yankunhuang-pku: OK, I modified it as required. I just don't think that's likely to happen

src/e2e-test/test-rule-functions.js
deepak-ms: please write it as "(s)". We don't need if/else for this.

src/e2e-test/test-rule-functions.js
deepak-ms: It might work for small errors, but doesn't seem sustainable as we add more stuff. For example, when we start adding new data types, there are always a lot of errors which get fixed slowly over a course of time. It would be great if we can add some tests in that period and improve on those test cases with time. we can think of it as code coverage numbers. 
How about you create a new user story for this change, which can be done separately?

src/e2e-test/test-rule-functions.js
yankunhuang-pku: OK, I have modified it.

src/e2e-test/test-rule-functions.js
yankunhuang-pku: Yes, I created a new story about Templates tests and updates using e2e test framework. I will try to use this rule and we can have more detailed discussion later.

PR #45, opened 2020-07-20
Personal/yankun/library packing configuration

yankunhuang-pku:
## Description

Export the conversion related function (`convert`) to pack a library, which will be used in the VS Code extension

## Testing

used `npm publish` to publish it and have used it in a VS Code extension, which can work effectively.


src/lib/workers/workerUtils.js
```diff
@@ -6,13 +6,16 @@
...
 module.exports.workerTaskProcessor = function (taskProcessor) {
-    parentPort.on('message', (msg) => {
-        taskProcessor(msg.msg)
-            .then(data => { 
-                msg.channelWorkerPort.postMessage(data);
-            }).catch(err => {
-                msg.channelWorkerPort.postMessage((err instanceof Error) ? err.message : err);
-            });
-    });
+    if (parentPort){
```
moria97, line 9: do we need handle else case here?

yankunhuang-pku: Good question~  @deepak-ms   Can you help us take a look at it? I'm not sure if I need to handle else case here, and just keep the old behavior.

yankunhuang-pku: I don't use the workerPool when packing this lib now. It seems that we don't need to handle else and when the parentPort is null, we just keep the old behavior and don't run the taskProcessor.

src/lib/index.js
```diff
@@ -0,0 +1,5 @@
+var convert = require('./workers/conversion').convert;
+
+module.exports = {
+    convert: convert
```
ACMoretxj, line 4: Maybe "{ convert }" could be simpler than "{ convert: convert }" if the external name is the same with the internal name.

BoyaWu10: Just curious why is this file named index.js?

ACMoretxj: > Just curious why is this file named index.js?

From my perspective, the index.js filename is a convention as a function hub of a lib. And there is a syntactic sugar that `require('@/package/lib')` is equal to `require('@/package/lib/index.js')`.

yankunhuang-pku: Thanks, I will change it "{ convert }".
I think, although the filename 'index.js' can be named arbitrarily, it generally defaults to index.js in node.js, which can be seen as the index of the exported variables and functions.

BoyaWu10: Got it. Thanks!

PR #62, opened 2020-08-22
2020.08.22 fixed nodejs ci error & refined groundtruth structure (by …

ACMoretxj:
## Description

Current consensus is that the `converter` project will run on windows platform, but the github workflow (`/,github/workflows/nodejs.yml`) wants a ubuntu environment, which may trigger unexpect exceptions. So I modified the platform from ubuntu to windows.

Besides above, the previous ground truth files contains unnecessary fields used for ui effections, such as `unusedSegments` and `invalidAccess`, I removed them for cleaner ground truth files content.


moria97:
For now, OSS Converter runs in linux containers, you can find more detail in the [Dockerfile](https://github.com/microsoft/FHIR-Converter/blob/master/Dockerfile). As node is a cross-platform runtime, what risks we may face after migrating to windows platform?
Agree with the ground truth part. However, do we have any test cases for ```unusedSegments ``` and ```invalidAccess```? These parts should be covered in additional testing for Web UI functions.

yankunhuang-pku:
Yes, I agree. However, if we separate them, we may still need to add some regression test for them separately because we will continue to use `unused segments` and `invalid access` in the future.

ACMoretxj:
Finally, I made it. There are no differet performace on desperate platforms any more.

src/e2e-test/regression-test/util/generate-ground-truths.js
```diff
@@ -16,7 +16,7 @@ const truthsExist = basePath => {
 
     const promises = [cdaPath, hl7v2Path].map(subPath => 
         new Promise(fulfill => fs.exists(subPath, fulfill)));
-    return Promise.all(promises).then(flags => flags.every(x => x));
+    return Promise.all(promises).then(flags => flags.some(x => x));
```
BoyaWu10, line 19: What does this change mean here?

ACMoretxj: The function is used to judge whether ground truth files have been generated. The existence of any part of ground truths will be regarded as the basis for not generating new ground truths. So `some` may be more appropriate than `every`.

yankunhuang-pku (approved):

src/e2e-test/regression-test/util/generate-ground-truths.js
BoyaWu10: Is it possible that only parts of the ground truth are generated last time?

src/e2e-test/regression-test/util/interceptors.js
```diff
@@ -35,17 +35,30 @@ class ExtraDynamicFieldInterceptor extends Interceptor {
...
         for (const entry of entries) {
             if ('resource' in entry && 'resourceType' in entry['resource']) {
                 if (entry['resource']['resourceType'] === 'DocumentReference') {
-                    entry['resource']['date'] = 'removed';
+                    const resource = entry['resource'];
+                    resource['date'] = 'removed';
+
+                    // The zlib.gzip result will be different on different platforms, see https://stackoverflow.com/questions/26516369/zlib-gzip-produces-different-results-for-same-input-on-different-oses.
+                    // Hence the hash result will be different too, which will trigger NodeJS CI error.
+                    if ('content' in resource && _.isArray(resource['content'])) {
```
moria97, line 50: Shall we keep consistent conversion result in different platforms?

src/e2e-test/regression-test/util/utils.spec.js
```diff
@@ -145,33 +143,33 @@ describe('Regression test testUtils - compareContent', () => {
...
     it ('should compare correctly when given parameters that only different with fields to be deleted', () => {
         const left = testData(), right = testData();
-        right.fhirResource.entry[0].resource.date = 'different';
+        right.entry[0].resource.date = 'different';
         assert.ok(utils.compareContent(...[left, right].map(JSON.stringify)));
     });
     it ('should throw error when given unequal plain object', () => {
         const left = testData(), right = testData();
-        right.fhirResource.entry[0].resource.fullUrl = 'Differen-Url';
+        right.entry[0].resource.fullUrl = 'Differen-Url';
```
moria97, line 151: typo Different

src/lib/handlebars-converter/handlebars-helpers.js
```diff
@@ -732,7 +732,8 @@ module.exports.external = [
         name: 'generateUUID',
         description: 'Generates a guid based on a URL: generateUUID url',
         func: function (urlNamespace) {
-            return uuidv3(''.concat(urlNamespace), uuidv3.URL);
+            const content = ''.concat(urlNamespace).replace(/(\r|\n|\r\n|\\r|\\n|\\r\\n)/gm, '');
```
moria97, line 735: Users may get different ID result after using this new code. We should create a bug and let Ranvijay know this change?

src/e2e-test/regression-test/util/generate-ground-truths.js
ACMoretxj: Of course, It's all about concret implementations, and this branch is just pick one of them.

src/lib/handlebars-converter/handlebars-helpers.js
ACMoretxj: Right, and an email has been sent to Ranvijay.

src/e2e-test/regression-test/util/utils.spec.js
ACMoretxj: Since the previous ground truths contain ui-related fields, there will be an extra `fhirResource`, currently we need not to do so because there are no ui-related fields and they will be test in other scenarioes.

src/e2e-test/regression-test/util/interceptors.js
ACMoretxj: Currently the difference will only impact regression test, and it will not it doesn't affect function(since the result encoded on Ubuntu can be correctly decoded on Windows).

src/e2e-test/regression-test/config/testcases-hl7v2.js
```diff
@@ -3,13 +3,18 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
-const dataFiles = [ 'ADT01-28.hl7' ];
-const templateFiles = [ 'ORU_R01.hbs' ];
+const patterns = [
+    { template: 'ADT_A01.hbs', data: [ 'ADT01-23.hl7', 'ADT01-28.hl7' ] },
```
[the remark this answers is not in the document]
ACMoretxj: I agree, I will correct it in future commits.

PR #73, opened 2020-09-22
Personal/xiatia/maintain main templates

ACMoretxj:
## Description

I have updated the HL7 main templates, primary concerns are as follows:
1. Only concentrate on the ID generation arguments;
2. If a field is not required according to the spec but used frequently in the template, add a "assert" statement for it(such as "PID-3-1");
3. If a field is required according to the spec but not used in the template, don’t handle it;
4. If a field is used to generate a GUID and it has no "assert" statement, add an "if" statement outside of it;

Besides these, I also updated the ground truths for regression tests, and added a ".gitattributes" file to handle whitespace changes of source files caused by platforms. And the "generateUUID" & "generateUUIDV2" have been updated to throw runtime errors when given undefined or null arguments.

There also is a **potential influence**, because the more stringent assert restriction, customers may encounter an error while passing a data file to a template which works normally before. ***We can also discuss this part in tomorrow's meeting***.

## Testing

All the modifications have been tested through regression tests and unittests.


src/templates/hl7v2/ADT_A01.hbs
```diff
@@ -3,50 +3,69 @@
...
-            {{>References/Procedure/subject.hbs ID=(generateUUID PR1) REF=(concat 'Patient/' (generateUUID PID-3-1))}},	
-            
+
+            {{#if PID-3-1}},
+                {{>Resources/Patient.hbs PID=PID PD1=PD1 NK1=NK1 ID=(generateUUID PID-3-1)}},
+            {{/if}},
+
+            {{#if PV1}},
+                {{>Resources/Encounter.hbs PV1=PV1 PV2=PV2 ID=(generateUUID PV1)}},
+                {{>References/Encounter/subject.hbs ID=(generateUUID PV1) REF=(concat 'Patient/' (generateUUID PID-3-1))}},
```
BoyaWu10, line 13: Are we checking the existence of REF (PID-3-1) here?

src/templates/hl7v2/OML_O21.hbs
```diff
@@ -3,83 +3,107 @@
     "type": "transaction",
     "entry": [
         {{#with (getFirstSegments msg.v2 'PID' 'PD1' 'PV1' 'PV2' 'ORC')}}
-            
-            {{assert ORC "Missing ORC"}},
```
BoyaWu10, line 7: Do I understand it correctly that instead of "assert" (check and throw exceptions if the field does not exist), we wrap an if condition outside it when the field is used?

src/e2e-test/regression-test/data/hl7v2/ADT_A01.hbs/ADT_A01.hbs-ADT01-23.hl7.json
```diff
@@ -173,20 +173,6 @@
...
-                "id": "4cfe8d6d-3fc8-3e41-b921-f204be18db31",
-                "subject": {
-                    "reference": "Patient/64bac34e-e611-3549-848b-89416176aa0b"
-                }
-            },
-            "request": {
-                "method": "POST",
-                "url": "Procedure/4cfe8d6d-3fc8-3e41-b921-f204be18db31"
-            }
-        },
```
BoyaWu10, line 189: Does the regression test work with both generateUUID and generateUUIDv2?

src/templates/hl7v2/ADT_A01.hbs
ACMoretxj: We only concentrate on the "ID" property, because the incorrection of "ID" means the whole resource is useless. But the incorrection of "REF" just means there is a single blank property and will not has such influences, just ignore it currently.

src/templates/hl7v2/OML_O21.hbs
ACMoretxj: Yes but partially, the "if" statements are used mainly to make sure there are no empty "ID" field(4cfe8d6d-3fc8-3e41-b921-f204be18db31), as well as doing part of what "assert" statement did.

The main function of "assert" statement will be replaced by a preprocessor(basically a V2 message validator) to execute more comprehensive restrictions.

src/e2e-test/regression-test/data/hl7v2/ADT_A01.hbs/ADT_A01.hbs-ADT01-23.hl7.json
ACMoretxj: Not exactly. The function of "generateUUIDV2" and that of "generateUUID" are different, we can't expect they will produce the same UUID. The plan is that if we decide to write(or automatically generate) new templates, we will use "generateUUIDV2" instead of "generateUUID", and hence the corresponding ground truths will be going to change.

src/templates/hl7v2/ADT_A01.hbs
ACMoretxj: Update: According to the suggestions from Ranvijay, the "REF" attribtues are also very important, so I will fixed it too.

src/templates/hl7v2/OML_O21.hbs
BoyaWu10: Got it. Thanks!

src/templates/hl7v2/ADT_A01.hbs
```diff
@@ -3,50 +3,99 @@
...
+            {{/if}},
+
             {{#with (getSegmentLists ../msg.v2 'NK1')}}
                 {{#each NK1 as |NK1Instance|}}
-    
-                    {{>Resources/RelatedPerson.hbs NK1=NK1Instance ID=(generateUUID NK1Instance)}},
-                    {{>References/RelatedPerson/patient.hbs ID=(generateUUID NK1Instance) REF=(concat 'Patient/' (generateUUID ../../PID-3-1))}},
-    
+
+                    {{#if NK1Instance}},
```
BoyaWu10, line 34: If NK1Instance is obtained from list (getSegmentLists), can we assume that NK1Instance will never be null or undefined?

ACMoretxj:
Many files are changed. BUT only line ending changes since the ".gitattributes" file, the new files have the standard line ending "\n" now.

src/templates/hl7v2/ADT_A01.hbs
```diff
@@ -3,50 +3,83 @@
...
-            {{>Resources/Procedure.hbs PR1=PR1 ID=(generateUUID PR1)}},
-            {{>References/Encounter/subject.hbs ID=(generateUUID PV1) REF=(concat 'Patient/' (generateUUID PID-3-1))}}, 
-            {{>References/Procedure/subject.hbs ID=(generateUUID PR1) REF=(concat 'Patient/' (generateUUID PID-3-1))}},	
-            
+
+            {{#if PID-3-1}},
+                {{>Resources/Patient.hbs PID=PID PD1=PD1 NK1=NK1 ID=(generateUUID PID-3-1)}},
+            {{/if}},
+
+            {{#if PV1}},
```
BoyaWu10, line 11: It seems PV1, PR1 comes from "getFirstSegments". Could we update "getFirstSegments" as well to avoid getting null or undefined?

src/templates/hl7v2/OML_O21.hbs
```diff
@@ -3,86 +3,110 @@
...
-    
+
+                    {{#if ../../PID-3-1}},
+                        {{>References/ServiceRequest/subject.hbs ID=(generateUUID (concat "ServiceRequest" ORCInstance)) REF=(concat 'Patient/' (generateUUID  ../../PID-3-1))}}
+                    {{/if}},
+
                     {{#with (getRelatedSegmentList ../../../msg.v2 'ORC' ORCInstance-1 'TQ1')}}
                         {{#each TQ1 as |TQ1Instance|}}
-                            
+
```
BoyaWu10, line 58: Same here for "getRelatedSegmentList". ("getSegmentLists", "getRelatedSegmentList" and "getFirstSegments" would better keep the same behavior.)

ACMoretxj:
Added a `.gitattributes` file to omit endless line ending issues. Please choose `Hide whitespace changes` while reviewing :)

src/templates/hl7v2/ADT_A01.hbs
ACMoretxj: Might a little different, 'cause this statement actually will not care about the return value from "getFirstSegments", it just declare a usage.

Even the "getFirstSegments" does not return the null or undefined value, the subsequent templates will also say that "I will use this segment, and I don't care whether it's undefined or not".

src/templates/hl7v2/OML_O21.hbs
ACMoretxj: Same reason with "ADT_A01".

BoyaWu10: getRelatedSegmentLists seems a bit different from getFirstSegments. It returns a list instead of objects, can we ensure the object in returned list is not null or empty (just like getSegmentLists)?

ACMoretxj: Sure we should also update it, I got a missunderstanding before .

BoyaWu10 (approved):

PR #77, opened 2020-09-30
Fix component overlap

BoyaWu10:
Current engine parses the whole string of a field into components. This could be wrong when field contains repetitions.

For example, the input segment is:
<pre>NK1|||||(130) 724-0433^PRN^PH^^^431^2780404~(330) 274-8214^ORN^PH^^^330^2748214</pre>
The engine parses NK1.5 into a list of components:
<pre>(130) 724-0433, PRN, PH, null, null, 431, <b>2780404~(330) 274-8214</b>, ORN, PH, null, null, 330, 2748214</pre>
The 7th component and ones after are wrong.

The fix is to split the field into repetitions first, and parse the **first** repetition as components, and all repetitions into `repeats`. Working item #76357 is created to track it.

### DIFF
--- a/src/lib/hl7v2/hl7v2.js
+++ b/src/lib/hl7v2/hl7v2.js
@@ -98,15 +98,14 @@ function parseHL7v2(msg) {
                     }
                     else {
                         var ocomps = CoverageArray.makeCoverageArray();
-
+                        var reps = fields[f].split(repetitionSeparator);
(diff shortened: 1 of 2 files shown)


BoyaWu10:
Close this PR according to offline discussion that we are going to fix this in templates by adding #each for fields with repeats.

Issue #60, opened 2020-08-07
Web UI and API return different json response


mikemassa84:
**Describe the bug**
The json response displayed in the Web UI contains a different bundle than the json response returned by the API for the same HL7v2 message. This can be confusing for users. The Web UI returns only the fhir bundle while the api returns the fhir bundle along with an invalidAccess object and an unusedSegments object.

**To Reproduce**
1. Go to the Web UI
2. Make sure HL7v2 is selected
3. Select ADT_A01.hbs from the template drop down.
4. Paste in an ADT message to convert in the top frame
5. Notice the bottom right frame contains only the converted fhir bundle
6. Send a POST request to /api/convert/hl7v2/ADT_A01.hbs with the same ADT message from step 4.
7. Notice the json response contains elements named "unusedSegments", "invalidAccess", and "fhirResource". This is different from the output of step 5.

**Expected behavior**
The response from the Web UI and the api should match.

**Screenshots**
Input and output are attached.

**Browser:**
Chrome

**Additional context**

[adt.txt](https://github.com/microsoft/FHIR-Converter/files/5042694/adt.txt)
[api_response.txt](https://github.com/microsoft/FHIR-Converter/files/5042695/api_response.txt)
[webui_response.txt](https://github.com/microsoft/FHIR-Converter/files/5042696/webui_response.txt)

ranvijaykumar:
The invalidAccess object and the unusedSegments are used by the UI for rendering purpose. We can take additional parameters to suppress those objects as those may not be useful in a pipeline. Would that work for you?

mikemassa84:
> The invalidAccess object and the unusedSegments are used by the UI for rendering purpose. We can take additional parameters to suppress those objects as those may not be useful in a pipeline. Would that work for you?

Yes, that would work for us.

ACMoretxj:
Close this issue, feel free to reactivate it for any questions.

PR #84, opened 2020-10-21
Personal/xiatia/additional parameter for api response

ACMoretxj:
## Description

* Added two extra parameters(`unusedSegments` and `invalidAccess`) in the query string for HL7V2 conversion for customizing response properties;
* Updated the documentation about API summary and data conversion;
* Fixed the document detail errors before: `/api/convert/{template}` -> `/api/convert/{srcDataType}/{template}`;

## Related issues

Addresses [issue #60 ].

## Testing

Added several extra unittests. All the UTs and E2E tests were passed.


docs/api-summary.md
```diff
@@ -23,5 +23,5 @@ The API is secured using an API Key.
...
-|POST      |/api/convert/{template}    |Takes data and converts to FHIR using the {template} that is stored on the server.|
+|POST      |/api/convert/{srcDataType} |Takes data, and temporary templates as input and outputs FHIR data after applying the templates on the data. The entry-point template is passed base64-encoded in templateBase64 parameter, whereas other overriding templates are passed in the templatesOverrideBase64 parameter.  templatesOverrideBase64 is a base64-encoded json object containing map between the template name and the template content. You can also set `unusedSegments` and `invalidAccess` parameters in the query string to fetch UI-related segments.|
+|POST      |/api/convert/{srcDataType}/{template}    |Takes data and converts to FHIR using the {template} that is stored on the server. You can also set `unusedSegments` and `invalidAccess` parameters in the query string to fetch UI-related segments.|
```
ranvijaykumar, line 27: {srcDataType} is part of the template name itself. Therefore, we need not include {srcDataType} here. See the output of {{converterbaseURL}}/api/templates to get the names of all the templates.

ACMoretxj: Got it, I'll remove the extra {srcDataType} path parameter.

docs/api-summary.md
```diff
@@ -23,5 +23,5 @@ The API is secured using an API Key.
...
-|POST      |/api/convert/{srcDataType} |Takes data, and temporary templates as input and outputs FHIR data after applying the templates on the data. The entry-point template is passed base64-encoded in templateBase64 parameter, whereas other overriding templates are passed in the templatesOverrideBase64 parameter.  templatesOverrideBase64 is a base64-encoded json object containing map between the template name and the template content.|
-|POST      |/api/convert/{template}    |Takes data and converts to FHIR using the {template} that is stored on the server.|
+|POST      |/api/convert/{srcDataType} |Takes data, and temporary templates as input and outputs FHIR data after applying the templates on the data. The entry-point template is passed base64-encoded in templateBase64 parameter, whereas other overriding templates are passed in the templatesOverrideBase64 parameter.  templatesOverrideBase64 is a base64-encoded json object containing map between the template name and the template content. You can also set `unusedSegments` and `invalidAccess` parameters in the query string to fetch UI-related segments.|
```
BoyaWu10, line 26: Will we request unusesSegments or invalidAccess independently? If not, we can combine them together into one option like "traceInfo=true".

BoyaWu10: The good point is that if we are going to add more trace or debug information in the future, we do not need to more "xxx=true" segments in the URL, and we can update the names of existing fields flexibly.

ACMoretxj: @ranvijaykumar What kind of implementation do you think is more appropriate? Use one parameter(like `traceInfo`) or multiple separate parameters(like `unusedSegments` and `invalidAccess` and so on)?

ranvijaykumar: unusedSegment is useful during template editing, while invalidAccess can be useful in production. Let's keep these separate.

BoyaWu10: Got it. Thanks!

BoyaWu10 (approved):

Issue #86, opened 2020-10-23
Git menu will trigger commit call on every click


ewoutkramer:
**Describe the bug**
When there are changes to commit, every single click on the Git dropdown menu will start a commit API call.

**To Reproduce**
1. Make some changes to the handlebar templates
2. Click on the "Git" menu
3. Watch the developer console to see the app invoke /commit on every click on the menu

**Expected behavior**
Only invoke this call when the "Commit changes" item is called

The callback for the commit is attached to the menu, instead of the "Commit changes" item.

ewoutkramer:
A PR will follow shortly.

ACMoretxj:
Hi Ewout, I can't reproduce the issue, it seems that the dropdown menu and the commit link(inside the dropdown menu) all work normally without functional interleaving on my machine. Could you please provide more detailed context information or screenshots?

ewoutkramer:
The event that triggers the call to commit seems to be attached to the element `#git-dropdown` - when I had commitable changes, every time I would just click the dropdown so it would open, I saw a commit being sent to the api endpoint.  So, I changed to code to be attached to the `<a>` link.  But I see now that the `on()` actually has the name of the child anchor there.  Hum.  

And I cannot reproduce it anymore, since I upgraded my node installation, and now all the Git functions don't work anymore anyway.  I'll try downgrading node and see if I can reproduce it again.

ewoutkramer:
Can't reproduce it myself anymore.  I'll close this until I hit this issue again ;-)

ACMoretxj:
Thank you Ewout, I will inform my teammates to pay attention to the GIT menu functions too, and please feel free to reactivate this issue for any questions :)

PR #87, opened 2020-10-23
fix: attaches the commitChanges() callback to the correct menu item

ewoutkramer:
## Description

This change attach the `commitChanges()` callback to the correct menu item, instead of the whole "Git" menu

## Related issues

Addresses #86.

## Testing

Watch the developer console and see that /commit is not called repeatedly for every single click on the menu.


ewoutkramer:
Can't reproduce it myself anymore.

PR #97, opened 2020-11-16
Personal/xiatia/new docker image options

ACMoretxj:
## Description

* Added the `latest` image tag while deploying.
* Use offline mode while validating using the official validator.

## Testing

All the unit tests are passed.

### DIFF
--- a/deploy/default-azuredeploy.json
+++ b/deploy/default-azuredeploy.json
@@ -36,9 +36,10 @@
       "type": "string",
       "allowedValues": [
         "v1.0.0",
-        "v2.0.0"
+        "v2.0.0",
+        "latest"
       ],
-      "defaultValue": "v2.0.0"
+      "defaultValue": "latest"
     }
   },
   "variables": {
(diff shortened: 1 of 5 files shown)


deploy/default-azuredeploy.json
```diff
@@ -36,9 +36,10 @@
       "type": "string",
       "allowedValues": [
         "v1.0.0",
-        "v2.0.0"
+        "v2.0.0",
+        "latest"
       ],
-      "defaultValue": "v2.0.0"
+      "defaultValue": "latest"
```
BoyaWu10, line 42: Does thee `latest` tag refer to the latest commit in master branch?

ACMoretxj: Yes.

BoyaWu10: I think we'd better use a stable version as default? It seems a bit risky to fetch latest commit from master as default.

BoyaWu10: Perhaps we could release another v2.0.1 or v2.1.0?

ACMoretxj: Sure, I believe a straightforward tag is more appropriate.

BoyaWu10 (approved):

PR #98, opened 2020-11-16
[DotLiquid] Add ID generation templates

BoyaWu10:
## Description

- Add new ID generation logics and templates.
- Move `SafeList`, `MemoryFileSystem` and `Evaluate` into `DotLiquids` folder
- Move redundant `with` in `include` tag

## Related issues

Addresses [issue #].

## Testing

Describe how this change was tested.


data/Templates/Hl7v2/ADT_A01.liquid
```diff
@@ -25,7 +27,7 @@
         {% endif -%}
 
         {% if firstSegments.PR1 -%}
-            {% assign pr1Id = firstSegments.PR1 | generate_uuid -%}
+            {% capture pr1Id -%}{% include 'ID/Procedure' with PR1: firstSegments.PR1, baseId: patientId -%}{% endcapture -%}
```
ACMoretxj, line 30: Maybe it's more consistent to keep ID related variables in the same naming style, always using the `${resourceType}Id` or `${v2SegmentType}Id`. If we choose the `resourceType`, the ID var name here should be `procedureId`, or `pr1Id` if we choose the later one style.

data/Templates/Hl7v2/ID/_RelatedPerson.liquid
```diff
@@ -0,0 +1,30 @@
+﻿{% if NK1 -%}
+{% if NK1.12.1 and NK1.12.4 -%}
+{% capture segment -%}{{NK1.12.1.Value}},{{NK1.12.4.Value}}{% endcapture -%}
+{% capture input -%}{{segment | generate_id_input: 'RelatedPerson', true, baseId}}{% endcapture -%}
+{% else -%}
+{% capture input -%}{{NK1.Value | generate_id_input: 'RelatedPerson', true, baseId}}{% endcapture -%}
+{% endif -%}
+{% elsif PID -%}
+{% if PID.21 -%}
+{% capture segment -%}{{PID.21.1.Value}},{{PID.21.4.Value}}{% endcapture -%}
```
ACMoretxj, line 10: We'd better also ensure the existence of `PID.21.1` and `PID.21.4` like `{% if NK1.12.1 and NK1.12.4 -%}` for consistency :)

BoyaWu10: Thanks for pointing this out. It is fixed in the update.

data/Templates/Hl7v2/ADT_A01.liquid
BoyaWu10: Agree. It is fixed in the update.

ACMoretxj (approved):
After solving the conflicts : )

ACMoretxj (approved):
A better version. Put the `generate_id_input` at the end of generating logic makes it much more clearer : )

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/GeneralFilters.cs
```diff
@@ -53,26 +53,37 @@ public static string Evaluate(string input, string property)
...
+                .Replace(encodingCharacters.RepetitionSeparator, ',');
+
+            formattedSegment = SpecialCharProcessor.Escape(Hl7v2EscapeSequenceProcessor.Unescape(formattedSegment, encodingCharacters));
+
+            return baseId != null ? $"{baseId}_{resourceType}_{formattedSegment}" : $"{resourceType}_{formattedSegment}";
         }
 
         public static string GenerateUUID(string input)
         {
             if (string.IsNullOrEmpty(input))
```
ACMoretxj, line 84: If the `input` is composed with whitespaces, we will still generate UUID for it, right?

BoyaWu10: Whitespaces will be transformed to null in `GenerateIdInput`, which is called before this function. But we can still check whitespaces here in case users call this function directly. I'll fix it in the update. Thanks!

ACMoretxj: Got it, thanks.

ACMoretxj (approved):

src/Microsoft.Health.Fhir.Liquid.Converter/Hl7v2/Models/Hl7v2Data.cs
```diff
@@ -10,12 +10,15 @@ namespace Microsoft.Health.Fhir.Liquid.Converter.Hl7v2.Models
 {
     public class Hl7v2Data : Drop
     {
-        public Hl7v2Data()
+        public Hl7v2Data(string value = null)
         {
+            Value = value;
```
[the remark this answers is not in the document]
BoyaWu10: This is used to store the whole message to compute the ID for Bundle.

data/Templates/Hl7v2/ID/_Device.liquid
```diff
@@ -0,0 +1,7 @@
+{% if HD.1 and HD.2 -%}
+{% capture segment -%}{{HD.1.Value}},{{HD.2.Value}}{% endcapture -%}
+{% capture input -%}{{segment | generate_id_input: 'Device', false}}{% endcapture -%}
```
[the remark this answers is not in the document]
BoyaWu10: Got it. I'll rename "segment" to "identifiers" in the update.

qiwjin (approved):

PR #119, opened 2020-12-06
Use default templates as embedded resource 

moria97:
## Description
Change default templates as embedded resource instead of content files, which embed the templates inside the dll and increases the flexibility to reference nuget packages.

## Related issues

Addresses [issue #].

## Testing

Run existing tests on template collection provider.


sowu880 (approved):

ACMoretxj:
@sowu880 Currently there are several tests failed in TemplateManagement subproject, it this in our expectations(like the tests are still under construction)?

src/Microsoft.Health.Fhir.TemplateManagement/Models/TemplateLayer.cs
```diff
@@ -17,21 +18,27 @@ public class TemplateLayer : OCIArtifactLayer
     {
         public Dictionary<string, Template> TemplateContent { get; set; }
 
-        public static TemplateLayer ReadFromFile(string filePath)
+        public static TemplateLayer ReadFromEmbeddedResource()
```
qiwjin, line 21: i feel we can keep the original api and also add a new one ReadFromStream().
both ReadFromFile and ReadFromEmbeddedResource will call ReadFromStream.

src/Microsoft.Health.Fhir.TemplateManagement/Microsoft.Health.Fhir.TemplateManagement.csproj
```diff
@@ -30,5 +30,15 @@
...
   </ItemGroup>
+  
+  <ItemGroup>
+    <EmbeddedResource Include="../../bin/DefaultTemplates.tar.gz">
+      <Link>DefaultTemplates.tar.gz</Link>
+    </EmbeddedResource>
+  </ItemGroup>
+
+  <Target Name="PreBuild" BeforeTargets="PreBuildEvent">
+    <Exec Command="mkdir ..\..\bin &amp; tar -zcvf ../../bin/DefaultTemplates.tar.gz -C ../../data/Templates/Hl7v2 ." />
```
qiwjin, line 41: nice, can you help to change the name into 'HL7v2DefaultTemplates'?

src/Microsoft.Health.Fhir.TemplateManagement/Models/TemplateLayer.cs
moria97: Sure. Have added the ReadFromFile() api back.

src/Microsoft.Health.Fhir.TemplateManagement/Microsoft.Health.Fhir.TemplateManagement.csproj
moria97: Maybe it's better to add a new "Renaming" PR as this name is referenced in multiple places @sowu880

src/Microsoft.Health.Fhir.TemplateManagement/Utilities/StreamUtility.cs
```diff
@@ -74,5 +74,16 @@ public static string CalculateDigestFromSha256(byte[] content)
...
+
+        public static string CalculateDigestFromSha256(Stream stream)
+        {
+            using SHA256 mySHA256 = SHA256.Create();
+            string hashedValue = "sha256:";
+            byte[] hashData = mySHA256.ComputeHash(stream);
+            string[] hashedStrings = hashData.Select(x => string.Format("{0,2:x2}", x)).ToArray();
+            hashedValue += string.Join(string.Empty, hashedStrings);
+            return hashedValue;
+        }
```
BoyaWu10, line 86: It seems these two functions share most logics. Could we convert the input stream to byte[] and call the original one?

moria97: Yeah, they are sharing same logic in get the hash value.  Converting to byte array will introduce extra cost than computing from a stream directly. I feel it's fine to keep two methods here.

BoyaWu10 (approved):

ACMoretxj (approved):

PR #127, opened 2020-12-11
Personal/sowu/cli exception 2.0

sowu880:
## Description

Refine exception handling in tool.
throw exceptions to program.cs and output error message in program.cs.


qiwjin (approved):

src/Microsoft.Health.Fhir.Liquid.Converter.Tool/Program.cs
```diff
@@ -14,19 +13,28 @@ namespace Microsoft.Health.Fhir.Liquid.Converter.Tool
...
+                parseResult.WithParsed<ConverterOptions>(options => ConverterLogicHandler.Convert(options));
+                await parseResult.WithParsedAsync<PullTemplateOptions>(options => TemplateManagementLogicHandler.PullAsync(options));
+                await parseResult.WithParsedAsync<PushTemplateOptions>(options => TemplateManagementLogicHandler.PushAsync(options));
+                parseResult.WithNotParsed((errors) => HandleOptionsParseError(parseResult));
+                return 0;
+            }
+            catch (Exception ex)
+            {
+                Console.Error.WriteLine($"Process failed: {ex.Message}");
+                return -1;
```
qiwjin, line 30: return 1?

sowu880: @yankunhuang-pku What about return 1 when process failed?

qiwjin:
please also help to add functional tests for it later.

src/Microsoft.Health.Fhir.Liquid.Converter.Tool/Program.cs
```diff
@@ -14,19 +13,28 @@ namespace Microsoft.Health.Fhir.Liquid.Converter.Tool
...
+            }
         }
 
         private static void HandleOptionsParseError(ParserResult<object> parseResult)
         {
             var usageText = HelpText.RenderUsageText(parseResult);
-            Console.WriteLine(usageText);
+            throw new InputParameterException(usageText);
```
BoyaWu10, line 37: Can we use `ArgumentException` instead of creating a new one?

BoyaWu10 (approved):

src/Microsoft.Health.Fhir.Liquid.Converter.Tool/Program.cs
yankunhuang-pku: As long as it is a non-zero number, I can identify it on my side.

yankunhuang-pku (commented):
Is it expected that both error messages will be output at the same time? Should the program be terminated after an error occurs?
`Fail to pull templates: The output folder is not empty. If force to override, please add -f in parameters`
`Process Exits: Fail to pull templates. Error: failed to authorize: failed to fetch anonymous token: unexpected status: 401 Unauthorized`

yankunhuang-pku (approved):

PR #138, opened 2020-12-16
[DotLiquid] Refine template loading performance

BoyaWu10:
When templates are loaded from local file system, the engine could load them gradually when they are used, instead of loading them at once from beginning. This helps improve the performance in VS Code extension.

### DIFF
--- a/src/Microsoft.Health.Fhir.Liquid.Converter/DotLiquids/MemoryFileSystem.cs
+++ b/src/Microsoft.Health.Fhir.Liquid.Converter/DotLiquids/MemoryFileSystem.cs
@@ -5,16 +5,23 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 using DotLiquid;
 using DotLiquid.FileSystems;
 using Microsoft.Health.Fhir.Liquid.Converter.Exceptions;
 using Microsoft.Health.Fhir.Liquid.Converter.Models;
+using Microsoft.Health.Fhir.Liquid.Converter.Utilities;
 
 namespace Microsoft.Health.Fhir.Liquid.Converter.DotLiquids
 {
     public class MemoryFileSystem : ITemplateFileSystem
     {
-        protected List<Dictionary<string, Template>> TemplateCollection { get; set; } = new List<Dictionary<string, Template>>();
+        protected const string TemplateFileExtension = ".liquid";
+
+        protected string TemplateDirectory { get; set; }
+
+        protected List<Dictionary<string, Template>> TemplateCollection { get; set; }
 
         public string ReadTemplateFile(Context context, string templateName)
         {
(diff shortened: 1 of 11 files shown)


src/Microsoft.Health.Fhir.Liquid.Converter/Hl7v2/Hl7v2TemplateProvider.cs
```diff
@@ -16,52 +15,34 @@ namespace Microsoft.Health.Fhir.Liquid.Converter.Hl7v2
...
+            if (!Directory.Exists(templateDirectory))
+            {
+                throw new ConverterInitializeException(FhirConverterErrorCode.TemplateFolderNotFound, string.Format(Resources.TemplateFolderNotFound, templateDirectory));
+            }
+
+            TemplateDirectory = templateDirectory;
+            TemplateCollection = LoadCodeSystemMapping();
         }
 
         public Hl7v2TemplateProvider(List<Dictionary<string, Template>> templateCollection)
```
ACMoretxj, line 29: How about deep-copying the `templateCollection` before it being assigned to `TemplateCollection`? Or the invoker might be able to manage the internal state under directly assigning mode.

BoyaWu10: Thanks, fixed in the update.

BoyaWu10: By offline discussion, we plan to copy list of dictionaries now.

src/Microsoft.Health.Fhir.Liquid.Converter/DotLiquids/MemoryFileSystem.cs
```diff
@@ -5,16 +5,23 @@
...
+using System.Linq;
 using DotLiquid;
 using DotLiquid.FileSystems;
 using Microsoft.Health.Fhir.Liquid.Converter.Exceptions;
 using Microsoft.Health.Fhir.Liquid.Converter.Models;
+using Microsoft.Health.Fhir.Liquid.Converter.Utilities;
 
 namespace Microsoft.Health.Fhir.Liquid.Converter.DotLiquids
 {
     public class MemoryFileSystem : ITemplateFileSystem
```
qiwjin, line 18: >MemoryFileSystem [](start = 17, length = 16)

suggest to add a new class 'TemplateLocalFileSystem'

BoyaWu10: Thanks, fixed in the update.

ACMoretxj (approved):

qiwjin (approved):

PR #67, opened 2020-09-03
Personal/xiatia/generate UUID v2

ACMoretxj:
## Description

Added "generateUUIDV2" helper function.

## Testing

See "handlebars-helper.spec.js".

PR #139, opened 2020-12-17
Updated main templates by adding some resources.

ACMoretxj:
* Added some resources to main templates.
* Updated related expected test resources for functional tests.
* Updated `Hl7v2DataParser` to handle MSH segments, added field separator as the second field of the MSH segment.

### DIFF
--- a/data/Templates/Hl7v2/ADT_A01.liquid
+++ b/data/Templates/Hl7v2/ADT_A01.liquid
@@ -10,6 +10,11 @@
             {% include 'Resource/Patient' PID: firstSegments.PID, PD1: firstSegments.PD1, NK1: firstSegments.NK1, ID: patientId -%}
         {% endif -%}
 
+        {% evaluate accountId using 'ID/Account' CX: firstSegments.PID.3 -%}
+        {% if accountId -%}
+            {% include 'Resource/Account' PID: firstSegments.PID, ID: accountId -%}
+        {% endif -%}
+
         {% evaluate encounterId using 'ID/Encounter' PV1: firstSegments.PV1, baseId: patientId -%}
         {% if encounterId -%}
             {% include 'Resource/Encounter' PV1: firstSegments.PV1, PV2: firstSegments.PV2, ID: encounterId -%}
--- a/data/Templates/Hl7v2/ADT_A01.liquid
+++ b/data/Templates/Hl7v2/ADT_A01.liquid
@@ -2,7 +2,7 @@
     "resourceType": "Bundle",
     "type": "transaction",
     "entry": [
-        {% assign firstSegments = hl7v2Data | get_first_segments: 'PID|PD1|PV1|PV2|PR1|AVR' -%}
+        {% assign firstSegments = hl7v2Data | get_first_segments: 'PID|PD1|PV1|PV2|PR1|AVR|MSH' -%}
         
         {% evaluate patientId using 'ID/Patient' PID: firstSegments.PID, type: 'First' -%}
         {% if patientId -%}
@@ -13,6 +13,11 @@
         {% evaluate accountId using 'ID/Account' CX: firstSegments.PID.3 -%}
         {% if accountId -%}
             {% include 'Resource/Account' PID: firstSegments.PID, ID: accountId -%}
+        {% endif -%}
+
+        {% evaluate organizationId using 'ID/Organization' HD: firstSegments.MSH.6 -%}
+        {% if organizationId -%}
+            {% include 'Resource/Organization' MSH: firstSegments.MSH, ID: organizationId -%}
         {% endif -%}
 
         {% evaluate encounterId using 'ID/Encounter' PV1: firstSegments.PV1, baseId: patientId -%}
@@ -50,6 +55,8 @@
         
         {% assign nk1SegmentLists = hl7v2Data | get_segment_lists: 'NK1' -%}
         {% for nk1Segment in nk1SegmentLists.NK1 -%}
+            {% include 'Resource/Patient' NK1: nk1Segment, ID: patientId -%}
+
             {% evaluate relatedPersonId using 'ID/RelatedPerson' NK1: nk1Segment, baseId: patientId -%}
             {% if relatedPersonId -%}
                 {% include 'Resource/RelatedPerson' NK1: nk1Segment, ID: relatedPersonId -%}
--- a/data/Templates/Hl7v2/ADT_A01.liquid
+++ b/data/Templates/Hl7v2/ADT_A01.liquid
@@ -3,6 +3,11 @@
     "type": "transaction",
     "entry": [
         {% assign firstSegments = hl7v2Data | get_first_segments: 'PID|PD1|PV1|PV2|PR1|AVR|MSH' -%}
+
+        {% evaluate messageHeaderId using 'ID/MessageHeader' MSH: firstSegments.MSH -%}
+        {% if messageHeaderId -%}
+            {% include 'Resource/MessageHeader' MSH: firstSegments.MSH, ID: messageHeaderID -%}
+        {% endif -%}
         
         {% evaluate patientId using 'ID/Patient' PID: firstSegments.PID, type: 'First' -%}
         {% if patientId -%}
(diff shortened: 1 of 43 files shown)


data/Templates/Hl7v2/Resource/_Location.liquid
```diff
@@ -7,17 +7,17 @@
...
-        {% include 'DataType/PL' PL: PR1.23 -%}
-        {% include 'DataType/PL' PL: PRT.9 -%}
-        {% include 'DataType/XONLocation' XON: ORC.21 -%}
+        {% include 'DataType/PL' PL: PR1.23.Repeats[0] -%}
+        {% include 'DataType/PL' PL: PRT.9.Repeats[0] -%}
+        {% include 'DataType/XONLocation' XON: ORC.21.Repeats[0] -%}
         "address":
         {
-            {% include 'DataType/XAD' XAD: ORC.22 -%}
+            {% include 'DataType/XAD' XAD: ORC.22.Repeats[0] -%}
```
qiwjin, line 15: consider to move repeats into XAD

ACMoretxj: Here we don't actually use the repeatable identifiers but just take the first element of the array, so the `ORC.22.Repeats[0] ` is just a single identifier, represents an unitary data.

data/Templates/Hl7v2/ADT_A01.liquid
```diff
@@ -2,14 +2,34 @@
...
+
+        {% evaluate provenanceId using 'ID/Provenance' MSH: firstSegments.MSH, baseID: patientId -%}
+        {% if provenanceId -%}
+            {% include 'Resource/Provenance' MSH: firstSegments.MSH, ORC: firstSegments.ORC, ID: provenanceId -%}
+        {% endif -%}
         
         {% evaluate patientId using 'ID/Patient' PID: firstSegments.PID, type: 'First' -%}
         {% if patientId -%}
             {% assign fullPatientId = patientId | prepend: 'Patient/' -%}
             {% include 'Resource/Patient' PID: firstSegments.PID, PD1: firstSegments.PD1, NK1: firstSegments.NK1, ID: patientId -%}
```
BoyaWu10, line 20: It seems we are using NK1 (firstSegments.NK1) when generating Patient resource, but NK1 is not passed as a argument here in `get_first_segments`. Could you please help to double check about it?

data/Templates/Hl7v2/ADT_A01.liquid
```diff
@@ -2,14 +2,34 @@
...
     "entry": [
-        {% assign firstSegments = hl7v2Data | get_first_segments: 'PID|PD1|PV1|PV2|PR1|AVR' -%}
+        {% assign firstSegments = hl7v2Data | get_first_segments: 'PID|PD1|PV1|PV2|PR1|AVR|MSH' -%}
+
+        {% evaluate messageHeaderId using 'ID/MessageHeader' MSH: firstSegments.MSH -%}
+        {% if messageHeaderId -%}
+            {% include 'Resource/MessageHeader' MSH: firstSegments.MSH, ID: messageHeaderID -%}
+        {% endif -%}
+
+        {% evaluate provenanceId using 'ID/Provenance' MSH: firstSegments.MSH, baseID: patientId -%}
```
BoyaWu10, line 12: We also need to update the variable in `ID/Provenance` from `ORC.Value` to `MSH.Value`. If we keep that template unchanged, `ID/Provenance` still expects ORC and will not be generated with MSH.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT01-23-expected.json
```diff
@@ -3,10 +3,40 @@
   "type": "transaction",
   "entry": [
     {
-      "fullUrl": "urn:uuid:64bac34e-e611-3549-848b-89416176aa0b",
+      "fullUrl": "urn:uuid:",
```
BoyaWu10, line 6: Can we keep ID in the groundtruth?

ACMoretxj: Sure, and I'll revert the latest commit since it will have conflict with another branch. In the following commits I'll add the ID.

data/Templates/Hl7v2/ADT_A01.liquid
ACMoretxj: I think the NK1 segments(the first one and the others) wil be injected in the following line #67 (`{% for nk1Segment in nk1SegmentLists.NK1 -%}`), so we may not need to handle the NK1 separately here.

data/Templates/Hl7v2/ADT_A01.liquid
ACMoretxj: Sure, I've updated the ID generation template.

data/Templates/Hl7v2/ADT_A01.liquid
BoyaWu10: Got it. If we are never using `firstSegments.NK1`, can we remove it from the arguments of Patient snippet?

ACMoretxj: Sure.

data/Templates/Hl7v2/ADT_A01.liquid
```diff
@@ -2,12 +2,32 @@
...
+        {% evaluate messageHeaderId using 'ID/MessageHeader' MSH: firstSegments.MSH -%}
+        {% if messageHeaderId -%}
+            {% include 'Resource/MessageHeader' MSH: firstSegments.MSH, ID: messageHeaderID -%}
+        {% endif -%}
         
         {% evaluate patientId using 'ID/Patient' PID: firstSegments.PID, type: 'First' -%}
         {% if patientId -%}
             {% assign fullPatientId = patientId | prepend: 'Patient/' -%}
-            {% include 'Resource/Patient' PID: firstSegments.PID, PD1: firstSegments.PD1, NK1: firstSegments.NK1, ID: patientId -%}
+            {% include 'Resource/Patient' PID: firstSegments.PID, PD1: firstSegments.PD1, ID: patientId -%}
```
ranvijaykumar, line 15: Why are we not using NL1 segment any more in the patient resource?

ranvijaykumar: I see that we are using it later, conditional to the presence of NK1 segment, which seems ok.

ranvijaykumar (approved):

data/Templates/Hl7v2/ADT_A01.liquid
ACMoretxj: Yes, we are using it in the following loop statements.

src/Microsoft.Health.Fhir.Liquid.Converter/Hl7v2/Hl7v2DataParser.cs
```diff
@@ -55,8 +55,20 @@ private List<Hl7v2Field> ParseFields(string dataString, Hl7v2EncodingCharacters
...
                 if (isHeaderSegment && f == 1)
                 {
+                    // field separator
+                    var fieldSeparatorComponents = new List<Hl7v2Component>
+                    {
+                        null,
+                        new Hl7v2Component(encodingCharacters.FieldSeparator.ToString(), new List<string> { null, encodingCharacters.FieldSeparator.ToString() }),
+                    };
+                    var fieldSeparatorField = new Hl7v2Field(encodingCharacters.FieldSeparator.ToString(), fieldSeparatorComponents);
+                    fields.Add(fieldSeparatorField);
```
BoyaWu10, line 69: Why do we need this here? It is already added in Line 78 of this file.

BoyaWu10: Solved by offline sync: https://hl7-definition.caristix.com/v2/HL7v2.8/Segments/MSH

PR #140, opened 2020-12-17
Support in linux and add functional tests

sowu880:
1. Add functional tests for vscode client in OCIArtifactFunctionaltests.cs. Simulate different situations and listed in the inline comments at the top of the cs file.
2. Update functional pipeline, refine registry building stage in windows vm.
3. Support template management CLI in linux system and could be extend to other system in the future.


src/Microsoft.Health.Fhir.TemplateManagement/Client/OrasClient.cs
```diff
@@ -93,5 +106,27 @@ private async Task OrasExecutionAsync(string command, string orasWorkingDirector
                 throw new OrasException(TemplateManagementErrorCode.OrasTimeOut, "Oras request timeout");
             }
         }
+
+        private void AddFilePermissionInLinuxSystem(string fileName)
```
BoyaWu10, line 110: Can we move this function to somewhere in common? It seems an useful function and used in unit tests as well.

src/Microsoft.Health.Fhir.TemplateManagement/Client/OrasClient.cs
```diff
@@ -18,11 +16,25 @@ namespace Microsoft.Health.Fhir.TemplateManagement.Client
     public class OrasClient : IOrasClient
     {
         private readonly string _imageReference;
+        private readonly string orasFileName;
```
BoyaWu10, line 19: Same naming styles for private fields would be better.

src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/OCIFileManagerTests.cs
```diff
@@ -18,28 +17,49 @@ public class OCIFileManagerTests
...
+            {
+                case PlatformID.Win32NT:
+                    _orasFileName = "oras-win.exe";
+                    break;
+                case PlatformID.Unix:
+                    _orasFileName = "oras-unix";
+                    OrasUtility.AddFilePermissionInLinuxSystem(_orasFileName);
+                    break;
+                default:
+                    throw new SystemException("System operation is not supported");
```
BoyaWu10, line 36: Is this exception caught outside into TemplateMangementException?

release.yml
```diff
@@ -127,7 +127,10 @@ stages:
         artifactName: FhirConverterBuild
         downloadPath: $(System.DefaultWorkingDirectory)
     - script: |
-        docker run --rm -d -p 5000:5000 --name registry stefanscherer/registry-windows:2.6.2
+        git clone -q https://github.com/sowu880/dockerfiles-windows.git
```
BoyaWu10, line 130: Are we going to maintain our own version of windows image?

BoyaWu10 (commented):
Let's update the title of the PR to a more formal one. The title will be included in release notes.

release.yml
sowu880: The public image can not be used for OCI artifacts.

sowu880:
Separate it into two PR #156 and #157

PR #146, opened 2020-12-18
Handle the repeatable identifiers

ACMoretxj:
* Handled part of repeatable(in V2 while not in FHIR) identifiers, only take the first non-empty element.

### DIFF
--- a/data/Templates/Hl7v2/Resource/_RelatedPerson.liquid
+++ b/data/Templates/Hl7v2/Resource/_RelatedPerson.liquid
@@ -34,12 +34,12 @@
         [
             { {% include 'DataType/XAD' XAD: ROL.11 -%} },
             {
-                {% include 'DataType/XAD' XAD: NK1.4 -%}
+                {% include 'DataType/XAD' XAD: NK1.4.Repeats[0] -%}
             },
             {
                 {% include 'DataType/XAD' XAD: NK1.32 -%}
             },
-            { {% include 'DataType/XAD' XAD: PRT.14 -%} },
+            { {% include 'DataType/XAD' XAD: PRT.14.Repeats[0] -%} },
         ],
         "telecom":
         [
@@ -64,11 +64,11 @@
             {
                 {% include 'DataType/XTN' XTN: NK1.41 -%}
             },
-            { {% include 'DataType/XTN' XTN: PRT.15 -%} },
+            { {% include 'DataType/XTN' XTN: PRT.15.Repeats[0] -%} },
         ],
         "name":
         [
-            { {% include 'DataType/XPN' XPN: NK1.2 -%} },
+            { {% include 'DataType/XPN' XPN: NK1.2.Repeats[0] -%} },
             { {% include 'DataType/XPN' XPN: NK1.30 -%} },
         ],
         "gender":"{{ NK1.15.Value | get_property: 'CodeSystem/Gender', 'code' }}",
@@ -82,7 +82,7 @@
                 },
             },
         ],
-        {% include 'DataType/XCNRelatedPerson' XCN: PRT.5 -%}
+        {% include 'DataType/XCNRelatedPerson' XCN: PRT.5.Repeats[0] -%}
     },
     "request":{
         "method":"PUT",
--- a/data/Templates/Hl7v2/Resource/_RelatedPerson.liquid
+++ b/data/Templates/Hl7v2/Resource/_RelatedPerson.liquid
@@ -34,12 +34,12 @@
         [
             { {% include 'DataType/XAD' XAD: ROL.11 -%} },
             {
-                {% include 'DataType/XAD' XAD: NK1.4.Repeats[0] -%}
+                {% include 'DataType/XAD' XAD: NK1.4.Repeats.0 -%}
             },
             {
                 {% include 'DataType/XAD' XAD: NK1.32 -%}
             },
-            { {% include 'DataType/XAD' XAD: PRT.14.Repeats[0] -%} },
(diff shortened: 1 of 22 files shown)


ranvijaykumar (approved):

BoyaWu10 (approved):
Let's make sure that we update the functional tests and make the same changes in template generation tool.

data/Templates/Hl7v2/Resource/_RelatedPerson.liquid
```diff
@@ -64,11 +64,11 @@
             {
                 {% include 'DataType/XTN' XTN: NK1.41 -%}
             },
-            { {% include 'DataType/XTN' XTN: PRT.15 -%} },
+            { {% include 'DataType/XTN' XTN: PRT.15.Repeats.0 -%} },
         ],
         "name":
         [
-            { {% include 'DataType/XPN' XPN: NK1.2 -%} },
+            { {% include 'DataType/XPN' XPN: NK1.2.Repeats.0 -%} },
```
BoyaWu10, line 71: Seen from the second commit of functional tests: We should use (a) `NK1.2.Repeats[0]` instead of (b) `NK1.2.Repeats.0` here. "`.`" operator always goes into a deeper level, e.g., from segment to field, from field to component, from component to subcomponent, so (b) refers to component NK1.2.0, which is always null (component starts from index 1).

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT01-23-expected.json
```diff
@@ -192,51 +192,33 @@
...
+        "url": "Location/9b2956e2-230b-9dd5-ad8d-0e697a8664c1"
       }
     },
     {
-      "fullUrl": "urn:uuid:a447a61d-cde0-394f-ab88-6a6f01215075",
+      "fullUrl": "urn:uuid:b8e7e176-84c6-7e2d-d422-a69df56229d8",
       "resource": {
         "resourceType": "RelatedPerson",
-        "id": "a447a61d-cde0-394f-ab88-6a6f01215075",
-        "address": [
```
BoyaWu10, line 203: We should be able to generate address and name of RelatedPerson. This should not be removed from groundtruth.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT01-23-expected.json
```diff
@@ -91,10 +91,10 @@
         "status": "unknown",
         "location": [
           {
-            "status": "active",
             "location": {
-              "reference": "Location/5f49ca69-cbda-3b30-8480-fa750dfee4a2"
-            }
+              "reference": "Location/"
```
BoyaWu10, line 95: This output is incorrect. It is caused by the `Encounter` template, which uses a Handlebars style to generate ID: `"reference":"Location/{{generateUUID PV1-3}}",`. Could you please help to create a working item about it?

data/Templates/Hl7v2/Resource/_RelatedPerson.liquid
ACMoretxj: Sure, I'll recover it.
The `.Repeats[0]` will cause crash while converting, and the `Repeats.0` works. I'll investigate the reason and try to fix it.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT01-23-expected.json
ACMoretxj: Sure, I'll create a task to fix this issue.

data/Templates/Hl7v2/Resource/_RelatedPerson.liquid
ACMoretxj: After changing from `.Repeats[0]` to `Repeats[0].Value`, the crashing disappeared. I'll update the repeatable identifiers in the main template fixing branch.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT01-23-expected.json
ACMoretxj: Solved by converting back `Repeats[0]` and udpating `Resource/Immunization`.

data/Templates/Hl7v2/Resource/_RelatedPerson.liquid
BoyaWu10: Yes, `Repeats[0]` is a Hl7v2Field object and `Repeats[0].Value` is a string. Which one to use depends on the scenario: When used in a `include` snippet, we usually pass a Hl7v2Field. If we want to render it directly, we usually use `Repeats[0].Value`.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT01-23-expected.json
BoyaWu10: Thanks!

PR #156, opened 2020-12-21
add functional tests for windows

sowu880:
Main changes:
1. Add functional tests for vscode client in OCIArtifactFunctionaltests.cs. Simulate different situations and listed in the inline comments at the top of the cs file.
2. Update functional pipeline, refine registry building stage in windows vm.

Useful Concept:
    Sequence number: 
       the sequence of each layer starts from 1. (e.g. sequence of base layer is 1, user layer is 2). If there is only one layer for an image, the sequence of the layer should be 1 or -1. (-1 mains the layer is not packed by our tool). If the image has multi layers, the sequence should be continuous and non-repeatable. (note! : Only layer with sequence 1 can be the base layer. )


release.yml
```diff
@@ -127,7 +127,10 @@ stages:
         artifactName: FhirConverterBuild
         downloadPath: $(System.DefaultWorkingDirectory)
     - script: |
-        docker run --rm -d -p 5000:5000 --name registry stefanscherer/registry-windows:2.6.2
+        git clone -q https://github.com/sowu880/dockerfiles-windows.git
+        cd dockerfiles-windows/registry/
+        docker build -t registry-windows:2.7.1 .
+        docker run --rm -d -p 5000:5000 --name registry registry-windows:2.7.1
```
ACMoretxj, line 133: Just for confirmation, if we run a docker in daemon mode, will the azure pipeline wait for the container to finish running?

src/Microsoft.Health.Fhir.TemplateManagement.FunctionalTests/OrasUtility.cs
```diff
@@ -0,0 +1,38 @@
...
+                StartInfo = new ProcessStartInfo(orasFileName),
+            };
+
+            process.StartInfo.Arguments = command;
+            process.StartInfo.RedirectStandardError = true;
+            process.EnableRaisingEvents = true;
+            process.Start();
+
+            StreamReader errStreamReader = process.StandardError;
+            process.WaitForExit(30000);
```
ACMoretxj, line 26: The constant value `30000` can be extracted out as a static readonly class field for unified management.

sowu880: Constants is an internal class could not be used here. Maybe hard code here is fine in test project?

ACMoretxj: I see. I believe the hard coding for tests is ok here.

release.yml
ACMoretxj: By teams discussion there is no issue for the docker-in-pipeline :)

ACMoretxj (approved):

release.yml
```diff
@@ -127,7 +127,10 @@ stages:
         artifactName: FhirConverterBuild
         downloadPath: $(System.DefaultWorkingDirectory)
     - script: |
-        docker run --rm -d -p 5000:5000 --name registry stefanscherer/registry-windows:2.6.2
+        git clone -q https://github.com/sowu880/dockerfiles-windows.git
```
moria97, line 130: nit: can we rename the repository to some meaningful name?

moria97 (approved):

PR #157, opened 2020-12-21
support linux

sowu880:
Support template management CLI in linux.


ACMoretxj (approved):

src/Microsoft.Health.Fhir.TemplateManagement/Utilities/OrasUtility.cs
```diff
@@ -0,0 +1,98 @@
...
+using Microsoft.Health.Fhir.TemplateManagement.Models;
+
+namespace Microsoft.Health.Fhir.TemplateManagement.Utilities
+{
+    public static class OrasUtility
+    {
+        public static string GetOrasFileName()
+        {
+            string orasFileName;
+            switch (Environment.OSVersion.Platform)
```
moria97, line 20: Maybe we can use [IsOSPlatform()](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.isosplatform?view=netcore-3.1) here?

src/Microsoft.Health.Fhir.TemplateManagement/Utilities/OrasUtility.cs
```diff
@@ -0,0 +1,98 @@
...
+                default:
+                    throw new TemplateManagementException("Operation system is not supported");
+            }
+
+            return orasFileName;
+        }
+
+        public static async Task OrasExecutionAsync(string command, string orasFileName, string orasWorkingDirectory)
+        {
+            TaskCompletionSource<bool> eventHandled = new TaskCompletionSource<bool>();
```
moria97, line 38: Can we remove the GetOrasFileName function and just check the platform information during each execution? The caller should not care about whether windows or unix binary is used.

moria97: Seems we don't need this utility? We can put this logic in the OrasClient

src/Microsoft.Health.Fhir.TemplateManagement/Utilities/OrasUtility.cs
```diff
@@ -0,0 +1,98 @@
...
+            else
+            {
+                throw new OrasException(TemplateManagementErrorCode.OrasTimeOut, "Oras request timeout");
+            }
+        }
+
+        public static void AddFilePermissionInLinuxSystem(string fileName)
+        {
+            var command = $"chmod +x {fileName}";
+            var escapedArgs = command.Replace("\"", "\\\"");
```
moria97, line 79: In which case will we have quotes in the command?

src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/OCIFileManagerTests.cs
```diff
@@ -18,28 +18,44 @@ public class OCIFileManagerTests
...
-            string command = $"push {_testOneLayerImageReference} {_baseLayerTemplatePath} {_userLayerTemplatePath}";
-            OrasExecution(command);
+            string command = $"push {_testMultiLayersImageReference} {_baseLayerTemplatePath} {_userLayerTemplatePath}";
+            try
+            {
+                await OrasUtility.OrasExecutionAsync(command, _orasFileName, Directory.GetCurrentDirectory());
+            }
+            catch
+            {
+                _isOrasValid = false;
```
moria97, line 57: Do we expect to skip or fail all tests when oras is invalid?

moria97 (changes requested):

sowu880:
> Shall we provide support for OSX (mac OS)? Seems a lot of people are using vs code on Mac

Yes, I think we could add a new task and realize in another PR, since we also need to setup OSX test environment in pipeline.

moria97 (approved):

PR #161, opened 2020-12-23
add image name validation

sowu880:
Add image name format validation: ^([a-z0-9])+((\.{1}|\-+|_{1,2})[a-z0-9]+)*$

### DIFF
--- a/src/Microsoft.Health.Fhir.TemplateManagement/Models/ImageInfo.cs
+++ b/src/Microsoft.Health.Fhir.TemplateManagement/Models/ImageInfo.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Health.Fhir.TemplateManagement.Models
         private const char ImageRegistryDelimiter = '/';
 
         // Reference docker's image name format: https://docs.docker.com/engine/reference/commandline/tag/#extended-description
-        private static readonly Regex s_imageNameRegex = new Regex(@"^[a-z0-9]+(([_\.]|_{2}|\-+)[a-z0-9]+)*(\/[a-z0-9]+(([_\.]|_{2}|\-+)[a-z0-9]+)*)*$");
+        private static readonly Regex ImageNameRegex = new Regex(@"^[a-z0-9]+(([_\.]|_{2}|\-+)[a-z0-9]+)*(\/[a-z0-9]+(([_\.]|_{2}|\-+)[a-z0-9]+)*)*$");
 
         public ImageInfo(string registry, string imageName, string tag = "latest", string digest = null)
         {
@@ -157,7 +157,7 @@ namespace Microsoft.Health.Fhir.TemplateManagement.Models
 
         private static void ValidateImageName(string imageName)
         {
-            if (!s_imageNameRegex.IsMatch(imageName))
+            if (!ImageNameRegex.IsMatch(imageName))
             {
                 throw new ImageReferenceException(TemplateManagementErrorCode.InvalidReference, @"Image name is invalid. Image name should contains lowercase letters, digits and separators. The valid format is ^[a-z0-9]+(([_\.]|_{2}|\-+)[a-z0-9]+)*(\/[a-z0-9]+(([_\.]|_{2}|\-+)[a-z0-9]+)*)*$");
             }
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.TemplateManagement/Models/ImageInfo.cs
```diff
@@ -55,35 +56,67 @@ public static bool IsDefaultTemplateImageReference(string imageReference)
...
+
+                ValidateImageName(imageName);
             }
+        }
 
-            return true;
+        private static void ValidateImageName(string imageName)
+        {
+            // Reference docker's image name format: https://docs.docker.com/engine/reference/commandline/tag/#extended-description
+            Regex imageNameRegex = new Regex(@"^([a-z0-9])+((\.{1}|\-+|_{1,2})[a-z0-9]+)*$");
```
moria97, line 115: We can make this regex a private/static member of this class to avoid creating object on each validation.
Besides, the regex can be ```^[a-z0-9]+([_\.|__|\-+][a-z0-9]+)*(\/[a-z0-9]+([_\.|__|\-+][a-z0-9]+)*)*$```
. [Reference](https://github.com/docker/distribution/blob/master/reference/reference.go)

moria97: Test cases:   
- org/image-name
   - org/image-name
    - image-name
     - image.name
    - image__name
    - image-----name

src/Microsoft.Health.Fhir.TemplateManagement/Models/ImageInfo.cs
```diff
@@ -55,35 +56,67 @@ public static bool IsDefaultTemplateImageReference(string imageReference)
...
+            }
+            catch
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        public static void ValidateImageReference(string imageReference)
```
moria97, line 72: Seems there is only one method calls ValidateImageReference, do we need to throw exceptions? Might be cleaner to just return false on fail? If we need the error message, I think we should throw an exception.

src/Microsoft.Health.Fhir.TemplateManagement/Models/ImageInfo.cs
```diff
@@ -55,35 +56,67 @@ public static bool IsDefaultTemplateImageReference(string imageReference)
         }
 
         public static bool IsValidImageReference(string imageReference)
+        {
+            try
+            {
+                ValidateImageReference(imageReference);
+            }
+            catch
```
moria97, line 64: This can be combined with the validateImageReference?

moria97 (changes requested):

src/Microsoft.Health.Fhir.TemplateManagement/Models/ImageInfo.cs
```diff
@@ -55,35 +59,57 @@ public static bool IsDefaultTemplateImageReference(string imageReference)
         }
 
         public static bool IsValidImageReference(string imageReference)
+        {
+            try
+            {
+                ValidateImageReference(imageReference);
```
moria97, line 65: move line 72 here.

src/Microsoft.Health.Fhir.TemplateManagement/Models/ImageInfo.cs
```diff
@@ -15,6 +16,9 @@ public class ImageInfo
         private const char ImageTagDelimiter = ':';
         private const char ImageRegistryDelimiter = '/';
 
+        // Reference docker's image name format: https://docs.docker.com/engine/reference/commandline/tag/#extended-description
+        private static readonly Regex _imageNameRegex = new Regex(@"^[a-z0-9]+(([_\.]|_{2}|\-+)[a-z0-9]+)*(\/[a-z0-9]+(([_\.]|_{2}|\-+)[a-z0-9]+)*)*$");
```
moria97, line 20: naming convention for static variable s_imageNameRegex. See [coding conventions](https://github.com/dotnet/corefx/blob/eb883d78defbc7d4cae3b8ebd0fa68852eb583e3/Documentation/coding-guidelines/coding-style.md)

moria97 (approved):

yankunhuang-pku (approved):

PR #166, opened 2020-12-24
Fix the handlebars-style reference ID in resource templates

ACMoretxj:
* All changed resource templates: `Encounter`, `PractitionerRole` and `Procedure`.
* All main templates are changed.
* Changed the ID from `generateUUID ...` to a ID variable passed by the invoker(usually the main templates).
* Changed the main templates to pass the referred ID.
* BTW, the reason why we directly set reference attributes inside resource templates instead of invoking `include Reference/.../...` is that if a resource has multiple reference attributes with the same type, there will be no way to tell them, like `PV1-3, Location.Reference` and `PV1-6, Location.Reference`.

### DIFF
--- a/data/Templates/Hl7v2/ADT_A01.liquid
+++ b/data/Templates/Hl7v2/ADT_A01.liquid
@@ -34,17 +34,17 @@
         {% if encounterId -%}
             {% include 'Resource/Encounter' PV1: firstSegments.PV1, PV2: firstSegments.PV2, ID: encounterId -%}
 
-            {% evaluate locationId using 'ID/Location' PL: firstSegments.PV1.3 -%}
-            {% if locationId -%}
-                {% include 'Resource/Location' PL: firstSegments.PV1.3, ID: locationId -%}
-                {% include 'Resource/Encounter' PV1: firstSegments.PV1, Location_ID_PV1_3: locationId, ID: encounterId -%}
+            {% evaluate locationId3 using 'ID/Location' PL: firstSegments.PV1.3 -%}
+            {% if locationId3 -%}
+                {% include 'Resource/Location' PL: firstSegments.PV1.3, ID: locationId3 -%}
             {% endif -%}
 
-            {% evaluate locationId using 'ID/Location' PL: firstSegments.PV1.6 -%}
-            {% if locationId -%}
-                {% include 'Resource/Location' PL: firstSegments.PV1.6, ID: locationId -%}
-                {% include 'Resource/Encounter' PV1: firstSegments.PV1, Location_ID_PV1_6: locationId, ID: encounterId -%}
+            {% evaluate locationId6 using 'ID/Location' PL: firstSegments.PV1.6 -%}
+            {% if locationId6 -%}
+                {% include 'Resource/Location' PL: firstSegments.PV1.6, ID: locationId6 -%}
             {% endif -%}
+
+            {% include 'Resource/Encounter' PV1: firstSegments.PV1, Location_ID_PV1_3: locationId3, Location_ID_PV1_6: locationId6, ID: encounterId -%}
 
             {% if patientId -%}
                 {% include 'Reference/Encounter/Subject' ID: encounterId, REF: fullPatientId -%}
--- a/data/Templates/Hl7v2/ADT_A01.liquid
+++ b/data/Templates/Hl7v2/ADT_A01.liquid
@@ -51,20 +51,22 @@
             {% endif -%}
         {% endif -%}
 
-        {% evaluate procedureId using 'ID/Procedure' PR1: firstSegments.PR1, baseId: patientId -%}
-        {% if procedureId -%}
-            {% include 'Resource/Procedure' PR1: firstSegments.PR1, ID: procedureId -%}
+        {% assign pr1SegmentLists = hl7v2Data | get_segment_lists: 'PR1' -%}
+        {% for pr1Segment in pr1SegmentLists.PR1 -%}
+            {% evaluate procedureId using 'ID/Procedure' PR1: pr1Segment, baseId: patientId -%}
+            {% if procedureId -%}
+                {% include 'Resource/Procedure' PR1: pr1Segment, ID: procedureId -%}
 
-            {% evaluate locationId using 'ID/Location' PL: firstSegments.PR1.23 -%}
-            {% if locationId -%}
-                {% include 'Resource/Location' PL: firstSegments.PR1.23, ID: locationId -%}
-                {% include 'Resource/Procedure' PR1: firstSegments.PR1, Location_ID_PR1_23: locationId, ID: procedureId -%}
-            {% endif -%}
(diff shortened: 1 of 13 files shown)


src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT04-23-expected.json
```diff
@@ -3,10 +3,10 @@
   "type": "transaction",
   "entry": [
     {
-      "fullUrl": "urn:uuid:18548d54-14a8-815a-c1bb-396f35b37164",
+      "fullUrl": "urn:uuid:",
```
BoyaWu10, line 6: Why do we remove id for this test case?

data/Templates/Hl7v2/Resource/_PractitionerRole.liquid
```diff
@@ -34,7 +34,7 @@
         [
             {
                 {% if ROL.13 -%}
-                "reference":"Location/{{generateUUID ROL-13}}",
+                "reference":"Location/{{ Location_ID_ROL_13 }}",
```
BoyaWu10, line 37: Currently PractitionerRole is never generated from our root templates, right?

data/Templates/Hl7v2/ADT_A01.liquid
```diff
@@ -37,35 +37,35 @@
...
                 {% include 'Reference/Encounter/Subject' ID: encounterId, REF: fullPatientId -%}
             {% endif -%}
         {% endif -%}
 
-        {% assign pr1SegmentLists = hl7v2Data | get_segment_lists: 'PR1' -%}
-        {% for pr1Segment in pr1SegmentLists.PR1 -%}
-            {% evaluate procedureId using 'ID/Procedure' PR1: pr1Segment, baseId: patientId -%}
-            {% if procedureId -%}
-                {% include 'Resource/Procedure' PR1: pr1Segment, ID: procedureId -%}
+        {% evaluate procedureId using 'ID/Procedure' PR1: firstSegments.PR1, baseId: patientId -%}
```
BoyaWu10, line 54: Why do we change it from segmentList to firstSegment?

data/Templates/Hl7v2/Resource/_PractitionerRole.liquid
ACMoretxj: Yes, currently we are not using it, we will update the main template to include more resources.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT04-23-expected.json
ACMoretxj: Might because of the copying issue, I'll fix it, thx.

data/Templates/Hl7v2/Resource/_PractitionerRole.liquid
BoyaWu10: Got it. Just to double confirm if I understand it correctly. Thanks!

data/Templates/Hl7v2/ADT_A01.liquid
ACMoretxj: I believe it's a mistake, I'll fix it.

data/Templates/Hl7v2/ADT_A01.liquid
```diff
@@ -2,7 +2,7 @@
     "resourceType": "Bundle",
     "type": "transaction",
     "entry": [
-        {% assign firstSegments = hl7v2Data | get_first_segments: 'PID|PD1|PV1|PV2|AVR|MSH' -%}
+        {% assign firstSegments = hl7v2Data | get_first_segments: 'PID|PD1|PV1|PV2|PR1|AVR|MSH' -%}
```
BoyaWu10, line 5: It seems we can remove the PR1 here.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT01-23-expected.json
```diff
@@ -196,6 +196,12 @@
               "reference": "Location/"
             },
             "status": "active"
+          },
+          {
+            "location": {
+              "reference": "Location/9b2956e2-230b-9dd5-ad8d-0e697a8664c1"
+            },
```
BoyaWu10, line 203: It seems we should replace the original one instead of adding a new one.

ACMoretxj: Yes, the blank reference is caused by a covert issue(the wrong condition of resource templates), currently it has been fixed.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT04-23-expected.json
```diff
@@ -234,15 +234,21 @@
         },
         "status": "unknown",
         "location": [
+          {
+            "status": "active"
+          },
+          {
+            "status": "completed"
+          },
```
BoyaWu10, line 242: How about these 2 status? They look redundant.

ACMoretxj: These two status are comming from different sources, and used to represent different Location status. In the test data we don't have adequate Location related segments, so there will be a single status here.

data/Templates/Hl7v2/ADT_A01.liquid
ACMoretxj: Sure, it's a redundant while solve conflicts.

data/Templates/Hl7v2/Resource/_Encounter.liquid
```diff
@@ -32,8 +32,8 @@
             {
                 "location":
                 {
-                    {% if PV1.6 -%}
-                    "reference":"Location/{{generateUUID PV1-6}}",
+                    {% if Location_ID_PV1_6 -%}
+                    "reference":"Location/{{ Location_ID_PV1_6 }}",
                     {% endif -%}
                 },
                 {% if PV1.6 -%}
```
BoyaWu10, line 39: I think we should also change the condition here?

BoyaWu10 (approved):
Let's create a PR in generation tool as well. :)

PR #168, opened 2020-12-25
Fixed issue that incorrectly handling of escaped characters in origin message

ACMoretxj:
**We are fixing the issue that causes conversion errors if the HL7 message contains escaped special characters.**

Here is the scenario, say we have a message in which a line is:
```hl7
PR1|1|FOO|FOO^ESCAPED ONE \T\ ESCAPED TWO^BAR|ESCAPED THREE \T\ ESCAPED FOUR|20201225000000||||||||||
```
The character `\T\` is used to escape `&` to prevent itself being cut to subcomponents. While we don't escape it back to `&` after converting, then the `PostProcessor` can't handle the `\T\` correctly because of the restriction of ANTLR rules, then error occurs.

This Pull-Request is meant to fix it.


src/Microsoft.Health.Fhir.Liquid.Converter/Hl7v2/Hl7v2Processor.cs
```diff
@@ -48,6 +49,8 @@ public string Convert(string data, string rootTemplate, ITemplateProvider templa
             var hl7v2Data = _dataParser.Parse(data);
             var context = CreateContext(templateProvider, hl7v2Data);
             var rawResult = RenderTemplates(template, context);
+            rawResult = Hl7v2EscapeSequenceProcessor.Unescape(rawResult, hl7v2Data.EncodingCharacters);
```
BoyaWu10, line 52: Can we do this in parsing instead of post-processing?

ACMoretxj: The special character escaping is used to represent semantical characters, if we parse them while parsing, they will be used to cut segments(or components) instead of being showed as text.

ACMoretxj: By offline discussion, I'll add this part in the parser.

src/Microsoft.Health.Fhir.Liquid.Converter/Hl7v2/Hl7v2DataParser.cs
```diff
@@ -151,5 +149,12 @@ private Hl7v2EncodingCharacters ParseHl7v2EncodingCharacters(string headerSegmen
                 SubcomponentSeparator = headerSegment[7],
             };
         }
+
+        private string SlackValue(string value, Hl7v2EncodingCharacters encodingCharacters)
```
BoyaWu10, line 153: Consider a better name?

BoyaWu10 (approved):

PR #170, opened 2021-01-06
Fixed trace info indexing issue for MSH segments while generating unused segments

ACMoretxj:
* MSH segments should be treated separately 'cause the first `|` is a special field, which caused incorrect indexing.
* Updated related test cases.

Well, we can't put the grammar escaping in post processing phase 'cause the `rawResult` contains tons of `"` to constrcut JSON result, the grammar escaping will mistakenly escape them, which causes error.
And we can't put the semantical unescaping in post processing phase 'cause the semantical unescaping must be done before grammar escaping, or some semantically escaped content(`\T\`, \E\) will be mistakenly grammar escaped(to `\\T\\`, `\\E\\`).

So we currently can only keep the semantical unescaping and grammar escaping in `NormalizeText`, and let the indexing offset by some spaces if the origin message contains semantically escapped characters(like `\T\`).

### DIFF
--- a/src/Microsoft.Health.Fhir.Liquid.Converter/Hl7v2/Models/Hl7v2TraceInfo.cs
+++ b/src/Microsoft.Health.Fhir.Liquid.Converter/Hl7v2/Models/Hl7v2TraceInfo.cs
@@ -81,7 +81,7 @@ namespace Microsoft.Health.Fhir.Liquid.Converter.Hl7v2.Models
 
         private static int FindOffsetInSegment(string segmentHeader, string segmentValue, Hl7v2EncodingCharacters encodingCharacters, int fieldIndex, int componentIndex)
         {
-            // All values($segmentValue and $fieldValue) need to be unescaped firstly (from "\\T" to "\T"),
+            // All values($segmentValue and $fieldValue) need to be unescaped firstly (from "\" to "\\"),
             // or the length will be incorrectly calculated
             segmentValue = SpecialCharProcessor.Unescape(segmentValue);
 
--- a/src/Microsoft.Health.Fhir.Liquid.Converter/Hl7v2/Models/Hl7v2TraceInfo.cs
+++ b/src/Microsoft.Health.Fhir.Liquid.Converter/Hl7v2/Models/Hl7v2TraceInfo.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections.Generic;
 using Microsoft.Health.Fhir.Liquid.Converter.Exceptions;
 using Microsoft.Health.Fhir.Liquid.Converter.Extensions;
+using Microsoft.Health.Fhir.Liquid.Converter.Hl7v2.InputProcessor;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Liquid.Converter/Hl7v2/Models/Hl7v2TraceInfo.cs
```diff
@@ -82,16 +79,26 @@ public static Hl7v2TraceInfo CreateTraceInfo(Hl7v2Data hl7v2Data)
             return new Hl7v2TraceInfo(unusedSegments);
         }
 
-        private static int FindOffsetInSegment(string segmentValue, Hl7v2EncodingCharacters encodingCharacters, int fieldIndex, int componentIndex)
+        private static int FindOffsetInSegment(string segmentHeader, string segmentValue, Hl7v2EncodingCharacters encodingCharacters, int fieldIndex, int componentIndex)
         {
+            // All values($segmentValue and $fieldValue) need to be unescaped firstly (from "\\T" to "\T"),
```
BoyaWu10: Let's change another example since \T\ is already replaced before.

src/Microsoft.Health.Fhir.Liquid.Converter/Hl7v2/Models/Hl7v2TraceInfo.cs
```diff
@@ -82,16 +79,26 @@ public static Hl7v2TraceInfo CreateTraceInfo(Hl7v2Data hl7v2Data)
...
 
-        private static int FindOffsetInSegment(string segmentValue, Hl7v2EncodingCharacters encodingCharacters, int fieldIndex, int componentIndex)
+        private static int FindOffsetInSegment(string segmentHeader, string segmentValue, Hl7v2EncodingCharacters encodingCharacters, int fieldIndex, int componentIndex)
         {
+            // All values($segmentValue and $fieldValue) need to be unescaped firstly (from "\\" back to "\"),
+            // or the length will be incorrectly calculated
+            segmentValue = SpecialCharProcessor.Unescape(segmentValue);
+
+            // MSH segment should be treated separately since the first '|' in MSH segment is a special field
+            if (segmentHeader == "MSH")
```
BoyaWu10, line 89: string.equals?

ACMoretxj: In this case `==` works equally as `Equals`, but I'll change it to `Equals` for consistency, thx :)

BoyaWu10: Not sure is segment ID with lower case valid? If so, let's set stringcomparison.invariantcultureignorecase.

BoyaWu10 (approved):

PR #174, opened 2021-01-07
Add reference properties for resource templates, add referred resources in main templates

ACMoretxj:
* Added reference attributes for resource template.
* Added referred resources in main templates.
* Please kindly ignore all the JSON changed since they are auto generated and are used for testing.

**The following ↓ is why we directly add reference properties in resource templates instead of using reference templates.**

---
image

### DIFF
--- a/data/Templates/Hl7v2/Resource/_Condition.liquid
+++ b/data/Templates/Hl7v2/Resource/_Condition.liquid
@@ -28,6 +28,12 @@
                 },
             ],
         },
+        "asserter":
+        {
+            {% if DG1.16 and Practitioner_ID_DG1_16 -%}
+            "reference":"Practitioner/{{ Practitioner_ID_DG1_16 }}",
+            {% endif -%}
+        },
     },
     "request":{
         "method":"PUT",
--- a/src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/OML_O21/MDHHS-OML-O21-1-expected.json
+++ b/src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/OML_O21/MDHHS-OML-O21-1-expected.json
@@ -1,6 +1,6 @@
 {
   "resourceType": "Bundle",
-  "type": "transaction",
+  "type": "batch",
   "entry": [
     {
       "fullUrl": "urn:uuid:3ec1a91c-4720-6f8c-d861-4fa35aff66a7",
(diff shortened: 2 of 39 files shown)


src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/OML_O21/MDHHS-OML-O21-1-expected.json
```diff
@@ -90,15 +138,15 @@
       }
     },
     {
-      "fullUrl": "urn:uuid:cec94dac-723d-33e4-3899-9e2b4a6c3be7",
+      "fullUrl": "urn:uuid:7bed24f6-d6fa-1218-ced1-b5d66ac0f024",
```
BoyaWu10, line 141: Why is the ID of provenance changed in tests?

ACMoretxj: Because the `Provenance` uses the `MSH.Value` for generating ID, and we added the `NormalizeText` recently, which changed the `MSH.Value`. Most other resources don' directly use the `MSH.Value`, so the IDs don't change.
Though ID changes, we can still say the `Provenance` resource is consistent.

BoyaWu10: Got it. Thanks for the investigation!

BoyaWu10 (approved):

data/Templates/Hl7v2/Resource/_Condition.liquid
```diff
@@ -28,6 +28,12 @@
                 },
             ],
         },
+        "asserter":
+        {
+            {% if DG1.16 and Practitioner_ID_DG1_16 -%}
```
BoyaWu10, line 33: Why do we need to check both of them? It seems only the latter one is used below.

ACMoretxj: Because we want to keep the reference properties consistent with other properties in the same level.
Like the `location.location.reference` and `location.status`.

BoyaWu10: By offline discussion, we are going to keep this for future refinement of conditions, like moving them to higher level of the tree. Besides, we need to track a item about using properties in references, like `Specimen.receivedTime/{{ Specimen_receivedTime_ID_OBR_13 }}` and `Practitioner|PractitionerRole/{{ Practitioner|PractitionerRole_ID_PV2_13 }}`.

PR #179, opened 2021-01-11
add dotnet restore

qiwjin:


release.yml
```diff
@@ -29,6 +29,13 @@ stages:
   - job: Build
     steps:
     - script: echo $(version)
+
+    - task: DotNetCoreCLI@2
+      inputs:
+        command: 'restore'
+        projects: '$(solution)'
+        feedsToUse: 'select'
```
BoyaWu10, line 37: Do we need to provide `vstsFeed` in the pipeline? In the [doc](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/dotnet-core-cli?view=azure-devops) it seems this variable is required:

#feedsToUse: 'select' # Options: select, config
#vstsFeed: # Required when feedsToUse == Select

BoyaWu10 (approved):

release.yml
qiwjin: do we have such feeds to restore or publish?

BoyaWu10: Currently the anonymizer/converter packages are published https://microsofthealthoss.visualstudio.com/FhirServer/_packaging?_a=feed&feed=Public (Here is the [nuget.config](https://github.com/microsoft/fhir-server/blob/main/nuget.config) of FHIR Server).

PR #192, opened 2021-01-26
Add custom dataHandlers

gjergjsheldija:
Since in healthcare we usually face the challenge to parse also different protocols then HL7v2 and CDA.
With this PR I've added the ability to add custom dataHanlders.
More in detail : 
- added custom dataHandlers, called parsers in this case
- added custom API entry to list them
- added dummy sample with tests
- changed UI to reflect this
- updated node git to v 0.25
- fixed nodegit tests
- updated code to run with nodejs > 10


ranvijaykumar:
Thanks for your PR. We are unable to accept this PR as our effort is currently focused on the Liquid branch, and we are taking only bug fixes in the Handlebars branch.

gjergjsheldija:
hi, 
Thank you for taking time to check it. 
Does it mean that the handlebars implementation is closed and no further development will happen in that version ?

ranvijaykumar:
At this point we do not have any new development planned for the handlebars version.

PR #199, opened 2021-02-01
Added FHIR parser functional tests.

ACMoretxj:
* Added FHIR parser functional tests, using all sample data as testing source.
* Removed `valueRange` property from `Resource/Observation` since it can't pass parser checking.
* Updated related test resources.

### DIFF
--- a/data/Templates/Hl7v2/Resource/_Observation.liquid
+++ b/data/Templates/Hl7v2/Resource/_Observation.liquid
@@ -58,24 +58,6 @@
                 {% endif -%}{% endif -%}
             },
         },
-        "valueRange":
-        {
-            {% if OBX.2.Value == "SN" -%}
-            {% include 'DataType/SNRange' SN: OBX.5 -%}
-            {% endif -%}
-            "low":
-            {
-                {% if OBX.2.Value == "SN" and OBX.5.3.Value == "-" -%}
-                {% include 'DataType/CWEQuantity' CWE: OBX.6 -%}
-                {% endif -%}
-            },
-            "high":
-            {
-                {% if OBX.2.Value == "SN" and OBX.5.3.Value == "-" -%}
-                {% include 'DataType/CWEQuantity' CWE: OBX.6 -%}
-                {% endif -%}
-            },
-        },
         "interpretation":
         [
             { {% include 'DataType/CWECodeableConcept' mapping: 'CodeSystem/InterpretationCode', CWE: OBX.8 -%} },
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Fhir.Liquid.Converter/Microsoft.Health.Fhir.Liquid.Converter.csproj
```diff
@@ -9,6 +9,7 @@
   <ItemGroup>
     <PackageReference Include="Antlr4.Runtime.Standard" Version="4.8.0" />
     <PackageReference Include="DotLiquid" Version="2.0.366" />
+    <PackageReference Include="Hl7.Fhir.R4" Version="2.0.3" />
```
moria97, line 12: Can we move the reference to the test project?

data/Templates/Hl7v2/Resource/_Observation.liquid
```diff
@@ -58,24 +58,6 @@
                 {% endif -%}{% endif -%}
             },
         },
-        "valueRange":
```
moria97, line 61: Will this deleted section cause "parse errors"?

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/RuleBasedTests.cs
```diff
@@ -140,6 +142,24 @@ public async Task CheckPassOfficialValidator(string templateName, string sampleP
             Directory.Delete(resultFolder, true);
         }
 
+        [Theory]
+        [MemberData(nameof(GetHL7V2Cases))]
+        [MemberData(nameof(GetCCDACases))]
+        public async Task CheckPassFhirParser(string templateName, string samplePath)
```
moria97, line 148: Better to add a sample that fails the parsing

BoyaWu10:
The reason it cannot pass the parser check seems that the value should be an integer (`182`) instead of a string (`"182"`). Can we try to fix the template instead of remove it?

ACMoretxj:
> The reason it cannot pass the parser check seems that the value should be an integer (`182`) instead of a string (`"182"`). Can we try to fix the template instead of remove it?

I've tried to do that before and it puts the folloing result, the weird point is the exception remains the same. So I decide to just remove the property currently and fix it later.
```json
"valueRange": {
      "low": {
            "value": 182
      }
},
```

data/Templates/Hl7v2/Resource/_Observation.liquid
ACMoretxj: Yes, or there will be a `{"While building a POCO: Element 'value' must not repeat (at Bundle.entry[5].resource[0].valueRange[0])"}` error.

src/Microsoft.Health.Fhir.Liquid.Converter/Microsoft.Health.Fhir.Liquid.Converter.csproj
ACMoretxj: Very reasonable, I'll do that.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/RuleBasedTests.cs
ACMoretxj: The original intention is to make sure our conversion result could pass the parser checking. If we add expect-fail-cases, it's more like we are testing the parser itself.
But in order ensure the basic functionality of the parser, I think it's also reasonable to add a basic expect-fail-case, I'll do that later.

moria97: Yes. Just want to make sure "Bad templates" will fail the test:)

BoyaWu10:
> > The reason it cannot pass the parser check seems that the value should be an integer (`182`) instead of a string (`"182"`). Can we try to fix the template instead of remove it?
> 
> I've tried to do that before and it puts the folloing result, the weird point is the exception remains the same. So I decide to just remove the property currently and fix it later.
> 
> ```json
> "valueRange": {
>       "low": {
>             "value": 182
>       }
> },
> ```

Sure. I saw the exception above and it is because `valueQuantity` and `valueRange` cannot appear at the same time, as defined in FHIR spec. We can fix it later.

BoyaWu10 (approved):

moria97 (approved):

PR #218, opened 2021-03-17
change name to ccda

hongyh13:
Unify the name to "C-CDA" in documents and "Ccda" in code


moria97:
Do we need to change the suffix of data files? ".cda" => ".ccda"

ranvijaykumar:
> Do we need to change the suffix of data files? ".cda" => ".ccda"

BoyaWu10:
> Do we need to change the suffix of data files? ".cda" => ".ccda"

I found some examples in the Hl7 [repo](https://github.com/HL7/C-CDA-Examples) and the file extension is "xml". Maybe we can use "xml" in our repo as well.

ranvijaykumar:
> > Do we need to change the suffix of data files? ".cda" => ".ccda"
> 
> I found some examples in the Hl7 [repo](https://github.com/HL7/C-CDA-Examples) and the file extension is "xml". Maybe we can use "xml" in our repo as well.

I think we will rely on the file extension to identify C-CDA documents in VS code extension. So, XML will not be ideal. We can change to ccda

hongyh13:
> > > Do we need to change the suffix of data files? ".cda" => ".ccda"
> > 
> > 
> > I found some examples in the Hl7 [repo](https://github.com/HL7/C-CDA-Examples) and the file extension is "xml". Maybe we can use "xml" in our repo as well.
> 
> I think we will rely on the file extension to identify C-CDA documents in VS code extension. So, XML will not be ideal. We can change to ccda

Thanks for the suggestion. I will change to ccda.

moria97 (approved):

release.yml
```diff
@@ -16,7 +16,7 @@ variables:
   functionalTests: "**/*FunctionalTests/*.csproj"
   buildConfiguration: 'Release'
   major: 3
-  minor: 4
+  minor: 5
```
BoyaWu10, line 19: Are we going to upgrade version to v3.5?

hongyh13: Yes, since cda default templates are supported, we could upgrade the version to v3.5. After discussing with Yue, I will undo it, and create a new pr to upgrade version.

BoyaWu10 (approved):

PR #222, opened 2021-03-25
Add support for time zone

BoyaWu10:

PR #224, opened 2021-03-26
Add support for time zone (#222)

BoyaWu10:
* Add support for time zone

### DIFF
--- a/src/Microsoft.Health.Fhir.Liquid.Converter/Filters/DateFilters.cs
+++ b/src/Microsoft.Health.Fhir.Liquid.Converter/Filters/DateFilters.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Health.Fhir.Liquid.Converter
             return ConvertDate(input, groups);
         }
 
-        public static string FormatAsDateTime(string input)
+        public static string FormatAsDateTime(string input, string timeZoneHandling = "local")
         {
             if (string.IsNullOrEmpty(input))
             {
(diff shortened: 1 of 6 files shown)


qiwjin (approved):

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/DateFilters.cs
```diff
@@ -48,28 +48,33 @@ public static string FormatAsDateTime(string input)
...
+            }
 
-                return result;
+            // Convert DateTime with time zone
+            var result = ConvertDateTime(input, groups);
+            if (groups["timeZone"].Success)
+            {
+                var timeZoneHour = int.Parse(groups["timeZoneHour"].Value);
+                var timeZoneMinute = int.Parse(groups["timeZoneMinute"].Value);
+                result += groups["sign"].Value + new TimeSpan(timeZoneHour, timeZoneMinute, 0).ToString(@"hh\:mm");
```
sowu880, line 62: Maybe we can put this logic into ConvertDateTime function and create datetime with TimeSpan in constructor ( rather than string concat)?
Than we don't need to TryParse again to validate result.

ACMoretxj (approved):

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/DateFilters.cs
BoyaWu10: Thanks for the comment! As this is a PR from dotliquid to main, we can see if we can refine it in later PRs. :)

PR #227, opened 2021-04-09
Add template version number and disclaimer into Provenance resource

ACMoretxj:
* According to previous discussion, we decide to add version number and disclaimer text in Provenance resource.
* Updated related testing resources.

### DIFF
--- a/data/Templates/Hl7v2/Resource/_Provenance.liquid
+++ b/data/Templates/Hl7v2/Resource/_Provenance.liquid
@@ -1,8 +1,19 @@
+{% assign generationInstant = "{{ "" | now }}" -%}
+{% assign templateVersion = "3.5.1" -%}
+{% assign rootTemplate = Root_Template -%}
+{% capture templateUrl -%}https://github.com/microsoft/FHIR-Converter/releases/download/v{{ templateVersion }}/Hl7v2DefaultTemplates.tar.gz{% endcapture -%}
+
 {
     "fullUrl":"urn:uuid:{{ ID }}",
     "resource":{
         "resourceType": "Provenance",
         "id":"{{ ID }}",
+        {% if Root_Template -%}
+        "text": {
+            "status": "generated",
+            "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\"><p><span>Resource bundle generated on {{ generationInstant }} using Microsoft FHIR Converter.</span><span>Template Version: {{ templateVersion }}.</span><span>Template URL: {{ templateUrl }}.</span><span>Root template: {{ rootTemplate }}.</span></p><p>Disclaimer: Real world HL7 messages vary in richness and level of conformance with the spec. The output of converter depends on the templates as well as the quality and richness of input messages. Therefore, it is important that you review and validate the Converter output before using those in production.</p></div>"
+        },
+        {% endif -%}
         "agent":
         [
             {
--- a/data/Templates/Hl7v2/Resource/_Provenance.liquid
+++ b/data/Templates/Hl7v2/Resource/_Provenance.liquid
@@ -8,7 +8,7 @@
         {% assign templateVersion = "3.5.1" -%}
         "text": {
             "status": "generated",
-            "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\"><p><span>Resource bundle generated on {{ generationInstant }} using Microsoft FHIR Converter.</span><span>Template Version: {{ templateVersion }}.</span><span>Template URL: https://github.com/microsoft/FHIR-Converter/releases/download/v{{ templateVersion }}/Hl7v2DefaultTemplates.tar.gz.</span><span>Root template: {{ Root_Template }}.</span></p><p>Disclaimer: Real world HL7 messages vary in richness and level of conformance with the spec. The output of converter depends on the templates as well as the quality and richness of input messages. Therefore, it is important that you review and validate the Converter output before using those in production.</p></div>"
+            "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\"><p><span>Resource bundle generated on {{ generationInstant }} using Microsoft FHIR Converter.</span><span>Template Version: {{ templateVersion }}.</span><span>Template URL: https://github.com/microsoft/FHIR-Converter/releases/download/v{{ templateVersion }}/Hl7v2DefaultTemplates.tar.gz.</span><span>Root template: {{ Root_Template }}.</span></p></div>"
         },
         {% endif -%}
         "agent":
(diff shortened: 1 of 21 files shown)


data/Templates/Hl7v2/Resource/_Provenance.liquid
```diff
@@ -1,8 +1,19 @@
+{% assign generationInstant = "{{ "" | now }}" -%}
+{% assign templateVersion = "3.5.1" -%}
+{% assign rootTemplate = Root_Template -%}
```
BoyaWu10, line 3: We can move them in `{% if Root_Template -%}` so we do not need to generate them for every Provenance. Also it looks that we can omit some variable assignment here. Maybe `templateVersion` would be enough.

ACMoretxj: I've removed `rootTemplate` and `templateUrl`, and move other variables into `IF` block.

BoyaWu10 (approved):

data/Templates/Hl7v2/Resource/_Provenance.liquid
```diff
@@ -3,6 +3,14 @@
     "resource":{
         "resourceType": "Provenance",
         "id":"{{ ID }}",
+        {% if Root_Template -%}
+        {% assign generationInstant = "{{ "" | now }}" -%}
+        {% assign templateVersion = "3.5.1" -%}
```
moria97, line 8: Is it possible to store version in a seperate file like config.json? Seems more friendly to maintain version upgrades.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT04-28-expected.json
```diff
@@ -226,6 +226,10 @@
       "resource": {
         "resourceType": "Provenance",
         "id": "a81dc37f-d7f3-c46a-00b1-4ff08fa99061",
+        "text": {
+          "status": "generated",
+          "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\"><p><span>Resource bundle generated on 2021-04-09T05:48:16.596Z using Microsoft FHIR Converter.</span><span>Template Version: 3.5.1.</span><span>Template URL: https://github.com/microsoft/FHIR-Converter/releases/download/v3.5.1/Hl7v2DefaultTemplates.tar.gz.</span><span>Root template: ADT_A01.</span></p><p>Disclaimer: Real world HL7 messages vary in richness and level of conformance with the spec. The output of converter depends on the templates as well as the quality and richness of input messages. Therefore, it is important that you review and validate the Converter output before using those in production.</p></div>"
```
moria97, line 231: Shall we wrap the version information to something like __ENGINE_VERSION__ in the test code like you do in the regex below so that we don't have to change all json files for each upgrade.
```csharp
            var regex = new Regex(@"Resource bundle generated on .* using Microsoft FHIR Converter");
```

BoyaWu10:
Can we add the same strategy for CCDA templates?

data/Templates/Hl7v2/Resource/_Provenance.liquid
ACMoretxj: The JSON file can't be directly used by template. We only need to change one variable to maintain the version number either in JSON file or in the Provenance template, and we only change this version number internally, so maybe the effort will be the same. Furthermore, the Provenance resource itself has a meta semantics, so that itself is as special as a "separate version template" and maybe we don't need an extra file with meta semantics.
If the HL7 and C-CDA can use shared templates in the future, I think it's a good idea to use a separate config file.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT04-28-expected.json
ACMoretxj: You mean we replace the version number `3.5.1` with `TEMPLATE_VERSION` literally in test JSON file so that we don't need to change the number each time we update the templates?

moria97: correct

ACMoretxj:
> Can we add the same strategy for CCDA templates?

UPDATE: By offline discussion we decide to add version number to C-CDA templates after the C-CDA templates are completed instead of in this pull-request.

data/Templates/Hl7v2/Resource/_Provenance.liquid
moria97: Yes. A seperate file would be better if we have multiple references. Yet I failed to find a solution to load an external variable from config files for templates.

moria97:
I noticed that the version number for our library (3.5.113.4 generated by azure pipeline) could be different from 3.5.1 in the template disclaimer. Shall we make the version numbers consistent?

ACMoretxj:
> I noticed that the version number for our library (3.5.113.4 generated by azure pipeline) could be different from 3.5.1 in the template disclaimer. Shall we make the version numbers consistent?

We want to keep the version of template consist with the release version. The detailed version `3.5.113.4` is generated by pipeline, we can't track and follow it.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ADT_A01/ADT04-28-expected.json
ACMoretxj: That's a good idea, I'll update the test resources.

moria97:
> > I noticed that the version number for our library (3.5.113.4 generated by azure pipeline) could be different from 3.5.1 in the template disclaimer. Shall we make the version numbers consistent?
> 
> We want to keep the version of template consist with the release version. The detailed version `3.5.113.4` is generated by pipeline, we can't track and follow it.

It's good to align the templates version with the release version. I think we should change nuget version to the release version too.  @BoyaWu10  thoughts?

moria97 (approved):

BoyaWu10:
> > > I noticed that the version number for our library (3.5.113.4 generated by azure pipeline) could be different from 3.5.1 in the template disclaimer. Shall we make the version numbers consistent?
> > 
> > 
> > We want to keep the version of template consist with the release version. The detailed version `3.5.113.4` is generated by pipeline, we can't track and follow it.
> 
> It's good to align the templates version with the release version. I think we should change nuget version to the release version too. @BoyaWu10 thoughts?

By offline discussion we are going to keep the major and minor versions the same for nuget version and release version.

PR #124, opened 2020-12-08
fix warning & refine tests

sowu880:
## Description

Describe the changes in this PR.

## Related issues

Addresses [issue #].

## Testing

Describe how this change was tested.

PR #232, opened 2021-04-21
Improve conversion performance by restricting template unfolding

ACMoretxj:
* Add input parameter judgement in all DataType templates;
* Add several judgement in main templates;


src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ORU_R01/LRI_2.0-NG_CBC_Typ_Message-expected.json
```diff
@@ -361,9 +351,6 @@
         "performer": [
           {
             "reference": "Organization/c8cd57a5-13a6-54d1-7dcd-d0a6b637c660"
-          },
-          {
-            "reference": "PractitionerRole/fbbaa96a-8a5d-f670-0fcf-838b21d4c6a8"
```
BoyaWu10, line 366: Just wondering why are we losing this reference?

data/Templates/Hl7v2/Resource/_Observation.liquid
```diff
@@ -125,9 +125,6 @@ Observation_Specimen_ID: A resource Id, used to fill "specimen.reference" proper
                 {% endif -%}
             },
             {
-                {% if OBX.23 and Organization_ID_OBX_23 -%}
-                "reference":"Organization/{{ Organization_ID_OBX_23 }}",
```
BoyaWu10, line 129: Are we going to remove this reference?

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Hl7v2/ORU_R01/LRI_2.0-NG_CBC_Typ_Message-expected.json
ACMoretxj: It's because of incorrect editing, I've fixed this issue, thanks!

data/Templates/Hl7v2/Resource/_Observation.liquid
ACMoretxj: Yes, this reference is duplicate with line #124, maybe it's caused by the error in mapping spreadsheets.

BoyaWu10: Got it. Thanks!

data/Templates/Hl7v2/ADT_A01.liquid
```diff
@@ -80,16 +80,24 @@
 
         {% for obxSegment in obxSegmentLists.OBX -%}
             {% evaluate organizationId_OBX_23 using 'ID/Organization' XON: obxSegment.23 -%}
-            {% include 'Resource/Organization' OBX: obxSegment, ID: organizationId_OBX_23 -%}
+            {% if obxSegment.24 or obxSegment.25 -%}
```
BoyaWu10, line 83: Can we put the checks in resources? It might a bit confusing to check the conditions in main templates as OBX.24 and OBX.25 are never mentioned in main templates.

ACMoretxj: It's mainly to avert such a scenario:
When the resource is included in another context without OBX segments, if there is a condition check about OBX inside the resource, the resource won't be rendered.

Take `Patient` for example, it will be included with `PID` and `NK1` separately, it's not appropriate to add either `PID` condition check or `NK1` condition check.

BoyaWu10 (approved):

PR #240, opened 2021-04-29
CCD templates

hongyh13:
Complete CCD templates derive from Handlebars templates.
Some previous issues are also fixed.


data/Templates/Ccda/Resource/_Patient.liquid
```diff
@@ -57,11 +57,13 @@
                         {
                             "system":"urn:ietf:bcp:47",
                             "code":"{{ languageCommunication.languageCode.code }}",
-                            "display":"{{ languageCommunication.languageCode.code | get_property: 'ValueSet/Language' }}",
+                            "display":"{{ languageCommunication.languageCode.code | downcase | get_property: 'ValueSet/Language' }}",
```
ACMoretxj, line 60: If the input parameter is null or empty, or even not string type, will `downcase` handle it appropriately?

ACMoretxj:
Let's also add C-CDA related rule-based tests in `RuleBasedTest.cs`. (It's also fine to do this in another pull-request).

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Ccda/CCD/CCD-expected.json
```diff
@@ -475,8 +475,8 @@
             "use": "mobile"
           },
           {
-            "system": "phone",
-            "value": "mailto://Isbella.Jones.CCD@gmail.com"
+            "system": "email",
+            "value": "//Isbella.Jones.CCD@gmail.com"
```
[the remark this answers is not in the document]
hongyh13: "mailto:" substring in "email" and "tel:" substring in "phone" are both removed for CCDA in handlebars and dotliquid.

data/Templates/Ccda/Resource/_Patient.liquid
hongyh13: I double-check and test it, it works correctly, no result will be generated for null or empty result, the original value will be converted to string for int, decimal or boolean type

hongyh13:
> Let's also add C-CDA related rule-based tests in `RuleBasedTest.cs`. (It's also fine to do this in another pull-request).

Thanks for the comment, I will add the tests in another pull-request.

data/Templates/Ccda/ValueSet/ValueSet.json
```diff
@@ -84,6 +90,9 @@
       "replaced": {
         "code": "amended"
       },
+      "ntered-in-error": {
```
BoyaWu10, line 93: entered-in-error?

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Ccda/CCD/CCD-expected.json
```diff
@@ -583,19 +583,6 @@
             }
           ]
         },
-        "reaction": [
```
BoyaWu10, line 586: Are we removing this property?

data/Templates/Ccda/ValueSet/ValueSet.json
BoyaWu10: Besides, are we making the same changes in ValueSets liquid templates?

hongyh13: typo, I will fix it.

hongyh13: > Besides, are we making the same changes in ValueSets liquid templates?

Yes, actually these changes are made to keep the same as ValueSets liquid templates.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Ccda/CCD/CCD-expected.json
hongyh13: according to https://www.hl7.org/fhir/allergyintolerance.html, there should have one manifestation in reaction, if there is not manifestation(entryRelationShip), I think we should also don't have "substance"

data/Templates/Ccda/ValueSet/_Language.liquid
```diff
@@ -1,26 +1,27 @@
...
+{% elsif downcaseCode == 'de' -%}
 "German",
-{% elsif code == 'da' -%}
+{% elsif downcaseCode == 'da' -%}
 "Danish",
-{% elsif code == 'fr' -%}
+{% elsif downcaseCode == 'fr' -%}
 "French",
 {% else -%}
 "{{ code }}",
```
qiwjin, line 27: shall we use case/when instead of if-elseif?

hongyh13: Sure, I will modify it.

data/Templates/Ccda/Resource/_Patient.liquid
ACMoretxj: Got it, thanks!

data/Templates/Ccda/ValueSet/_Language.liquid
qiwjin: can you also help to compare the performance between "if" and "case"?

hongyh13: I compare the performance based on our sample data and a big dataset, there is no obvious difference in performance between "if" and "case". I think it may be because the conditional selection itself does not take up much time.

BoyaWu10 (approved):

ACMoretxj (approved):

PR #249, opened 2021-05-13
Update consistent indentions of templates

ACMoretxj:
**Only template modification**

* Update indentions of ID templates;
* Update indentions of DataType templates;
* Update indentions of Resource templates;
* CodeSystem templates might be removed later, we don't update them this PR;


data/Templates/Hl7v2/DataType/_EIPIdentifier.liquid
```diff
@@ -1,26 +1,24 @@
...
-            "system":"http://terminology.hl7.org/CodeSystem/v2-0203",
-            {% endif -%}
-        },
-    ],
-},
-
-{% endif -%}
\ No newline at end of file
+    "value":"{{ EIP.1.1 }}",
+    "value":"{{ EIP.2.1 }}",
```
ranvijaykumar, line 3: Please review this. Why do we have two value nodes here? EIP.1 and EIP.2 are different entities. Do we need an If condition here?

ranvijaykumar (approved):

data/Templates/Hl7v2/DataType/_EIPIdentifier.liquid
ACMoretxj: The mapping spreadsheet `EIP[Identifier]` contains two 'identifier' properties, which points to `EI[Identifier]` template and will be unfolded to `value: ...`. The condition statements make little difference, there are same situations for the following `code` and `system` properties.

I'm not pretty sure how to update this template, do you have any suggestions?

BoyaWu10 (approved):

PR #252, opened 2021-05-19
complete C-CDA templates comparable with handlebars

hongyh13:


data/Templates/Ccda/HistoryandPhysical.liquid
```diff
@@ -0,0 +1,22 @@
+{
```
BoyaWu10, line 1: Can we add some functional tests for these main template as well, just like the ones we do for HL7 v2 (ADT, ORU, OML, VXU)?

hongyh13: Sure, I added them.

data/Templates/Ccda/Entry/AllergiesAndAdverseReaction/_entry_act_entryRelationship.liquid
```diff
@@ -1,4 +1,4 @@
-{% if relationship.observation.templateId | to_json_string | contains: '2.16.840.1.113883.10.20.22.4.7' -%}
+{% if relationship.observation.templateId | to_json_string | contains: '"2.16.840.1.113883.10.20.22.4.7"' -%}
```
ACMoretxj, line 1: What's the consideration that we add double quotation marks here?

hongyh13: If we don't add a double quotation here, a string contains this substring may also be included. For example, we could like to match "2.16.840.1.113883.10.20.22.4.7" here, if we don't add a double quotation, another id "2.16.840.1.113883.10.20.22.4.71"  will also match this condition.

ACMoretxj: Got it, we have to use double quotation since we are using `contains` filter.

ACMoretxj (approved):

BoyaWu10 (approved):

PR #256, opened 2021-05-26
Fix CI/CD issue caused by action server's performance

ACMoretxj:


deploy/default-azuredeploy.json
```diff
@@ -34,17 +34,11 @@
     },
     "imageTag": {
       "type": "string",
-      "allowedValues": [
-        "v1.0.0",
-        "v2.0.0",
-        "v2.1.0"
-      ],
-      "defaultValue": "v2.1.0"
+      "defaultValue": "healthplatformregistry.azurecr.io/fhirconverter:v2.1.0"
```
BoyaWu10, line 37: If we are storing the complete one here, maybe it does not belong to `imageTag`?

test/e2e-test/rule-test/util/test-rule-functions.spec.js
```diff
@@ -240,7 +240,7 @@ describe('testRule', function () {
         const reqJson = null;
         const resJson = JSON.parse(fs.readFileSync(path.join(__dirname, '../test-samples/FHIR-R4/sample3-right.json')));
         assert.ok(testRules.officialValidator(reqJson, resJson).valid);
-    }).timeout(60000);
+    }).timeout(180 * 1000);
```
BoyaWu10, line 243: Does the issue caused by the timeout settings?

ACMoretxj: Yes, the performance of github action server is not that stable, so the time usage of validating phase will fluctuate.
I change the timeout limit to a relatively big number to eliminate the influence of performance fluctuation.

deploy/default-azuredeploy.json
ACMoretxj: Absolutely, the `imageTag` is not a good name,  I'll change it to `imagePath`.

test/e2e-test/rule-test/util/test-rule-functions.spec.js
BoyaWu10: Got it. Thanks!

BoyaWu10 (approved):

qiwjin (approved):

Issue #257, opened 2021-05-26
NodeJs Version


castasint:
Hi - I see that there is a constraint on the NodeJs version when using "engines": { "node": ">=10.10.0 <11.0" },. On 30th of last month(04/30/2021) was the end-of-life for any 10.x versions. So, is there any plan of upgrading the project to v14 or v12(LTS) so it can be safely used? Please advise.

Also, if there is a guide that already explains how the project can be upgraded, please point me towards it. Happy to answer any further questions.

![image](https://user-images.githubusercontent.com/16588875/119666358-edbb4180-be52-11eb-9bb1-7785c7f29c69.png)

I've added this in the discussion forum a few days back[https://github.com/microsoft/FHIR-Converter/discussions/237](https://github.com/microsoft/FHIR-Converter/discussions/237). But, now I think it should be called as a potential issue.
Thanks,
Hanuma

PR #265, opened 2021-06-30
Upgrade nodejs to 14 for handlebars

hongyh13:
Upgrade NodeJs to [14](https://nodejs.org/download/release/v14.17.1/)

The end-of-life date of NodeJs version of 10.x was 2021/04/30, so upgrade NodeJs version to the LTS version 14.17.1. 
Antlr and nodegit are also upgraded to support NodeJs 14.

[discussion reference](https://github.com/microsoft/FHIR-Converter/discussions/237)
[issue reference](https://github.com/microsoft/FHIR-Converter/issues/257)


BoyaWu10 (approved):

src/lib/outputProcessor/autogen/json.interp
```diff
@@ -0,0 +1,40 @@
+token literal names:
```
ACMoretxj, line 1: These auto generated files will be created at runtime, and they are unable to be fine-grained controlled. Maybe it's more appropriate to add them in .gitignore.

package.json
```diff
@@ -84,33 +84,33 @@
...
     "fhir": "^4.7.9",
     "fs-extra": "^8.1.0",
-    "handlebars": "^4.7.6",
-    "lodash": "^4.17.19",
+    "handlebars": "^4.7.7",
+    "lodash": "^4.17.21",
     "memory-cache": "^0.2.0",
     "ncp": "^2.0.0",
-    "nodegit": "^0.24.3",
+    "nodegit": "^0.27.0",
```
ACMoretxj, line 108: We can just remove all the nodegit related snippets since we are not using the git mechanism any more. Previously we just hided the entrance of git functions, now we can thoroughly remove the git functions.

src/lib/outputProcessor/autogen/json.interp
hongyh13: Thanks for the suggestion, I will do it.

package.json
hongyh13: I will check with @qiwjin: if we remove nodegit in this pr or in another pr?

ACMoretxj: Whether to remove the nodegit or not is not an emergent thing, I think we can remove it later instead of in this pull request.

hongyh13: I will remove it in another pr.

ACMoretxj (approved):

PR #273, opened 2021-07-12
Upgrade handlebars version to 2.1.1

hongyh13:
Upgrade handlebars version to 2.1.1.

### DIFF
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -1,6 +1,9 @@
 name: Docker Image CI
 
 on: 
+  pull-request:
+    branches:
+      - handlebars
   push:
     branches:
       - handlebars
--- a/.github/workflows/dockerimage.yml
+++ b/.github/workflows/dockerimage.yml
@@ -1,7 +1,7 @@
 name: Docker Image CI
 
 on: 
-  release:
+  push:
     branches:
       - handlebars
   schedule:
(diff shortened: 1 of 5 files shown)


.github/workflows/dockerimage.yml
```diff
@@ -27,9 +30,14 @@ jobs:
       id: package-version
       uses: martinbeentjes/npm-get-version-action@v1.0.0
 
+    - uses: actions/checkout@v2
+      with:
+        ref: v${{ steps.package-version.outputs.current-version }}
```
ACMoretxj, line 35: Since there are two repos here, it's better to speficy their paths separately to avoid confusion, and specify working directory in subsequent steps:

```yml
- name: Checkout master repo
    uses: actions/checkout@v2
    with:
      repository: $REPO
      ref: ...
      path: ./master

- name: Checkout release repo
    uses: actions/checkout@v2
    with:
      repository: $REPO
      ref: ...
      path: ./release

- name: Subsequent steps
    working-directory: ./master OR ./release
    run: .....
```

.github/workflows/dockerimage.yml
```diff
@@ -16,7 +19,7 @@ jobs:
     steps:
     - name: 'Checkout Github Action'
       uses: actions/checkout@master
-      
+
```
ACMoretxj, line 22: It's not common to checkout the same repo multiple times, we can add comments here to explain why we are using this somehow weired workaround.

hongyh13: I remove the second code checkout now, and this workflow is triggered only when a new release is published.

hongyh13:
I test the built image, it works.

BoyaWu10 (approved):

ACMoretxj (approved):

PR #275, opened 2021-07-15
Add initial support for converting JSON to FHIR

BoyaWu10:
Add initial support for converting JSON to FHIR.
The JSON conversion template includes:
- an example template to convert a simple generic JSON to FHIR Patient
- an example template to convert FHIR STU3 ChargeItem to R4

Parsers, processors and project structures are also refined.


src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Json/SimplePatient-expected.json
```diff
@@ -0,0 +1,44 @@
+{
```
hongyh13, line 1: To keep the same structure as `HL7v2` and `C-CDA`, how about adding a subfolder for these data, such as `TestData/Expected/Json/Sample/SimplePatient-expected.json`?

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests.csproj
```diff
@@ -36,6 +36,12 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup>
```
hongyh13, line 39: If a subfolder is added for expected JSON output, the added code is no longer needed.

data/Templates/Json/SimplePatient.liquid
```diff
@@ -0,0 +1,54 @@
+{
+    "resourceType": "Patient",
+    "id": "{{ msg.PatientId | to_json_string | generate_uuid }}",
```
hongyh13, line 3: I'm just curious the key `msg` is used here the same as `C-CDA`, so will we use `msg` for all the datatypes except `HL7v2`?

BoyaWu10: Yes, I think we can use `msg` for all data types except `hl7v2` (due to legacy compatibility) to keep consistency.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/TestData/Expected/Json/SimplePatient-expected.json
BoyaWu10: Sure, subfolders are created according to the root templates they use.

src/Microsoft.Health.Fhir.Liquid.Converter/Json/DictionaryJsonConverter.cs
```diff
@@ -5,10 +5,9 @@
 
 using System;
```
hongyh13, line 6: This file seems a base class, and `C-CDA` inherits it, is it appropriate to put it here?

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/ProcessorTests.cs
```diff
@@ -62,12 +67,17 @@ public static IEnumerable<object[]> GetValidInputsWithProcessSettings()
             yield return new object[]
             {
                 new Hl7v2Processor(null), new Hl7v2Processor(new ProcessorSettings()), new Hl7v2Processor(positiveTimeOutSettings), new Hl7v2Processor(negativeTimeOutSettings),
-                new Hl7v2TemplateProvider("TestTemplates"), _hl7v2TestData,
+                new Hl7v2TemplateProvider("TestTemplates"), _hl7v2TestData
```
hongyh13, line 70: How about using a constant string since it appears many times?

BoyaWu10: Sure. A constant is added to replace the raw strings.

src/Microsoft.Health.Fhir.Liquid.Converter/Json/DictionaryJsonConverter.cs
BoyaWu10: `DictionaryJsonConverter` is the base class of `XmlDictionaryJsonConverter`. The former one is used in JSON-to-FHIR to parse generic JSON data to dictionary directly. Based on that, the latter one has some extra logics to parse the XML-converted JSON to work with the CCDA templates. I've moved `DictionaryJsonConverter` to namespace `Converter` instead of `Converter.Json` in the update.

src/Microsoft.Health.Fhir.Liquid.Converter/Ccda/CcdaDataParser.cs
```diff
@@ -18,7 +18,7 @@ public static class CcdaDataParser
     {
         public static IDictionary<string, object> Parse(string document)
         {
-            if (string.IsNullOrEmpty(document))
+            if (string.IsNullOrWhiteSpace(document))
```
ACMoretxj, line 21: The error code `NullOrEmptyInput` is used to handle `IsNullOrEmpty` and `IsNullOrWhiteSpace` simultaneously, maybe we can split the error code to two types or rename it to a more general name.

src/Microsoft.Health.Fhir.Liquid.Converter/Constants.cs
```diff
@@ -9,5 +9,6 @@ public static class Constants
     {
         public const string Hl7v2DataKey = "hl7v2Data";
```
moria97, line 10: Curiously, do we intend to use "msg" for all resource types except v2?

src/Microsoft.Health.Fhir.Liquid.Converter/Ccda/CcdaDataParser.cs
BoyaWu10: Thanks, the error code and message are unified to `NullOrWhiteSpaceInput` to handle `string.IsNullOrWhiteSpace()` only.

src/Microsoft.Health.Fhir.Liquid.Converter/DictionaryJsonConverter.cs
```diff
@@ -41,11 +40,9 @@ private object ReadValue(JsonReader reader)
                     return ReadObject(reader);
                 case JsonToken.StartArray:
                     return ReadArray(reader);
-                case JsonToken.String:
-                    // Remove line breaks to avoid invalid line breaks in json value
-                    return Regex.Replace(reader.Value.ToString(), @"\r\n?|\n", string.Empty);
```
moria97, line 46: Is it ok to remove this line ?

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonDataParser.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public class JsonDataParser
+    {
+        public IDictionary<string, object> Parse(string json)
+        {
+            try
+            {
+                if (string.IsNullOrWhiteSpace(json))
+                {
+                    throw new DataParseException(FhirConverterErrorCode.NullOrEmptyInput, Resources.NullOrEmptyInput);
```
moria97: Seems we can put this throw outside of try?

src/Microsoft.Health.Fhir.Liquid.Converter/Models/DataType.cs
```diff
@@ -9,5 +9,6 @@ public enum DataType
     {
         Hl7v2,
         Ccda,
+        Json,
```
moria97, line 12: Shall we add comments for all data types? Looks like we are going to have a long list of data types in the future.

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonProcessor.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class JsonProcessor : BaseProcessor
+    {
+        private readonly JsonDataParser _dataParser = new JsonDataParser();
+
+        public JsonProcessor(ProcessorSettings processorSettings = null)
+            : base(processorSettings)
+        {
+        }
+
+        public override string Convert(string data, string rootTemplate, ITemplateProvider templateProvider, TraceInfo traceInfo = null)
```
moria97, line 22: Looks like we can reuse the Convert logic for v2, ccda and json processors?

moria97: How about add a new interface with a cancellation token for rendering

src/Microsoft.Health.Fhir.Liquid.Converter/DictionaryJsonConverter.cs
BoyaWu10: Yes, line breaks are valid in XML but invalid in generic JSON. The logic of removing line breaks is now placed in `XmlDictionaryJsonConverter`, which is used for parsing C-CDA XML documents.

src/Microsoft.Health.Fhir.Liquid.Converter/Models/DataType.cs
BoyaWu10: Thanks, I've added comments for this class to indicate that this is the input data types.

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonProcessor.cs
BoyaWu10: > Looks like we can reuse the Convert logic for v2, ccda and json processors?

The conversion logics for HL7 v2, C-CDA and JSON are a bit different. HL7 v2 and C-CDA to FHIR conversion need to load different CodeSystem/ValueSet mappings in their context. HL7 v2 supports returning the debug information (unused segments, etc.) in the output. So the `Convert` function has different implementations for different input data types.

BoyaWu10: > How about add a new interface with a cancellation token for rendering

Thanks, this is already supported in `BaseProcessor`.

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonDataParser.cs
BoyaWu10: Yes! I forgot to move this one out.

data/Templates/Json/SimplePatient.liquid
```diff
@@ -0,0 +1,54 @@
+{
```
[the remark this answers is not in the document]
moria97: Got it. What about ```ExamplePatient``` since it's an example?

src/Microsoft.Health.Fhir.Liquid.Converter/Constants.cs
moria97: What about adding ```msg```  key support for hl7v2 and keep compatible with the original "hl7v2Data". So old clients can keep using the "hl7v2Data" but new clients should use the consistent "msg" key.

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonProcessor.cs
moria97: > > Looks like we can reuse the Convert logic for v2, ccda and json processors?
> 
> The conversion logics for HL7 v2, C-CDA and JSON are a bit different. HL7 v2 and C-CDA to FHIR conversion need to load different CodeSystem/ValueSet mappings in their context. HL7 v2 supports returning the debug information (unused segments, etc.) in the output. So the `Convert` function has different implementations for different input data types.

Yes. Since context and traceinfo are referenced in all interfaces. Looks we can keep a basic function that create context and do trace info collections in BaseProcessor and override the functions when needed (like in v2 and cda processor)?

moria97: > > How about add a new interface with a cancellation token for rendering

Do we plan to add cancellation token in the core convert logic as well as in the interface? like [here](https://github.com/dotliquid/dotliquid/pull/406/files)

data/Templates/Json/SimplePatient.liquid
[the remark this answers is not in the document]
BoyaWu10: Sure, this would be a better name.

src/Microsoft.Health.Fhir.Liquid.Converter/Constants.cs
moria97: After offline discussion, we decide not to change the key for v2 as it is widely used in templates and it's difficult to change the existing templates.

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonProcessor.cs
BoyaWu10: Thanks for the suggestion! A base function is added in `BaseProcessor` . `BaseProcessor` would provide virtual methods `CreateContext` and `CreateTraceInfo` for derived processors to override.

As for cancellation token, we could write a new `CreateContext` for it when we update the dotliquid package.

moria97: Sounds good. Thanks!

moria97 (approved):
:shipit:

hongyh13 (approved):

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonDataParser.cs
```diff
@@ -0,0 +1,38 @@
...
+        public IDictionary<string, object> Parse(string json)
+        {
+            if (string.IsNullOrWhiteSpace(json))
+            {
+                throw new DataParseException(FhirConverterErrorCode.NullOrWhiteSpaceInput, Resources.NullOrWhiteSpaceInput);
+            }
+
+            try
+            {
+                var data = JsonConvert.DeserializeObject<object>(json, new JsonSerializerSettings { ContractResolver = new DictionaryContractResolver() });
```
qiwjin, line 25: >DeserializeObject [](start = 39, length = 17)

how about using JObject.Parse?

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonDataParser.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Health.Fhir.Liquid.Converter.Exceptions;
+using Microsoft.Health.Fhir.Liquid.Converter.Models;
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.Fhir.Liquid.Converter.Json
+{
+    public class JsonDataParser
```
qiwjin, line 14: >JsonDataParser [](start = 17, length = 14)

can we have a shared interface?

ACMoretxj (approved):

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonDataParser.cs
BoyaWu10: Sure. According to offline discussion we'll do several refinements for code structure. I'll do it in another PR.

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonDataParser.cs
BoyaWu10: JObject.Parse does not accept `ContractResolver` as input. Here we need `DictionaryContractResolver` to parse input data to `IDictionary<string, object>` instead of `JObject` /`JArray`, which is not accepted by dotliquid.

src/Microsoft.Health.Fhir.Liquid.Converter/BaseProcessor.cs
```diff
@@ -49,6 +51,36 @@ protected virtual Context CreateContext(ITemplateProvider templateProvider, IDic
             return context;
         }
 
+        protected virtual void CreateTraceInfo(IDictionary<string, object> data, TraceInfo traceInfo)
```
qiwjin, line 54: >virtual [](start = 18, length = 7)

change to abstract?

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonDataParser.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public IDictionary<string, object> Parse(string json)
+        {
+            if (string.IsNullOrWhiteSpace(json))
```
qiwjin, line 18: >IsNullOrWhiteSpace [](start = 23, length = 18)

shall we allow empty?

src/Microsoft.Health.Fhir.Liquid.Converter/BaseProcessor.cs
BoyaWu10: I think virtual might be more suitable here. The default behavior is no action. HL7 v2 overrides this and C-CDA and JSON just use the default. If we change it to abstract, all processors will need to implement their own `CreateTraceInfo`.

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonDataParser.cs
BoyaWu10: Currently null, empty or string consists only of white-space characters are not allowed. Do we have any scenario that may accept empty input?

src/Microsoft.Health.Fhir.Liquid.Converter/Json/JsonDataParser.cs
BoyaWu10: `DeserializeObject` is replaced with `JToken.Parse` and `JTokenExtensions` in the update.

qiwjin (approved):

PR #277, opened 2021-07-15
add rule based test for C-CDA

hongyh13:


src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/RuleBasedTests.cs
```diff
@@ -57,30 +63,103 @@ public static IEnumerable<object[]> GetHL7V2Cases()
...
+            var result = await ConvertData(templateName, samplePath, dataType);
             var patients = result.SelectTokens("$.entry[?(@.resource.resourceType == 'Patient')].resource.id");
             Assert.Equal(1, patients?.Count());
         }
 
+        [Theory]
+        [MemberData(nameof(GetCcdaCases))]
+        public async Task CheckReferenceResourceId(string templateName, string samplePath, DataType dataType)
+        {
+            Dictionary<string, string> resourceIdMapping = new Dictionary<string, string>();
```
BoyaWu10, line 137: It looks we can use a List<string> to store resource IDs like `{resourceType}/{resourceId}`. Resource IDs only need to be unique within a resource type. For example, we can have `Patient/example` and `Observation/example` at the same time.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/RuleBasedTests.cs
```diff
@@ -57,30 +63,103 @@ public static IEnumerable<object[]> GetHL7V2Cases()
...
+                var resourceId = resource.SelectTokens("$.id").First().ToString();
+                var resouceType = resource.SelectTokens("$.resourceType").First().ToString();
+                Assert.False(resourceIdMapping.ContainsKey(resourceId));
+                resourceIdMapping.Add(resourceId, resouceType);
+            }
+
+            // check reference resouce id exists
+            foreach (var resource in resources)
+            {
+                RevealObject(resource, 0, resourceIdMapping);
```
BoyaWu10, line 153: Maybe we can have a more specific name for it like `RevealReferences`?

hongyh13: fixed.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/RuleBasedTests.cs
hongyh13: HashSet is used since the time complexity of `List.Contains` is O(N).

qiwjin (approved):

BoyaWu10 (approved):

PR #279, opened 2021-07-15
Removing other OS references to avoid confusion

ginalee-dotcom:


BoyaWu10:
It seems we have already merged a PR with the same change in the main branch: https://github.com/microsoft/FHIR-Converter/pull/258.

PR #280, opened 2021-07-15
Create partial-template-concept.md on dotliquid branch

ginalee-dotcom:


BoyaWu10:
There is a similar document in dotliquid engine [here](https://github.com/microsoft/FHIR-Converter/blob/main/docs/SnippetConcept.md), but it only contains the contents for Hl7 v2. (`Partial template` in handlebars is called `snippet` in dotliquid.) It seems we can update the C-CDA contents there instead of creating a new one.

PR #292, opened 2021-08-03
Adds dependabot config

brendankowitz:


qiwjin:
@brendankowitz  We have enable Dependabot alerts for vulnerabilities that affect our dependencies.  for non-vulnerable dependencies upgrade, we might want to use different update strategy.

qiwjin:
close inactive pr

PR #307, opened 2021-08-26
Fix issues for mac

sowu880:
1. Fix the path of Oras exe.
2. Add checking exit code of Oras process.
3. Add oras_osx_amd64.tar.gz in published package.


src/Microsoft.Health.Fhir.TemplateManagement/Client/OrasClient.cs
```diff
@@ -105,9 +105,9 @@ public static async Task OrasExecutionAsync(string command, string orasWorkingDi
             if (process.HasExited)
             {
                 string error = errStreamReader.ReadToEnd();
-                if (!string.IsNullOrEmpty(error) || process.ExitCode != 0)
+                if (process.ExitCode != 0)
                 {
-                    throw new OrasException(TemplateManagementErrorCode.OrasProcessFailed, "Oras process failed" + error);
+                    throw new OrasException(TemplateManagementErrorCode.OrasProcessFailed, $"Oras process failed. {errStreamReader.ReadToEnd()}");
```
BoyaWu10, line 110: Can we use the `error` variable here?

BoyaWu10 (approved):

PR #323, opened 2021-10-08
Xoriant/m1 template removed

v-shnarang:
We have removed the below ADT message template and updated the Bundle type to "batch"
o	ADT_A11: Cancel admit/visit notification 
o	ADT_A13: Cancel discharge/end visit
o	ADT_A25: Cancel pending discharge
o	ADT_A26: Cancel pending transfer 
o	ADT_A27: cancel pending admit 
o	ADT_A29: Delete person information
o	ADT_A45: Move visit information – visit number 

We have added the creation of bundle ID from MSH.10 and reference the same in provenance resource.


BoyaWu10:
> We have removed the below ADT message template and updated the Bundle type to "batch" o ADT_A11: Cancel admit/visit notification o ADT_A13: Cancel discharge/end visit o ADT_A25: Cancel pending discharge o ADT_A26: Cancel pending transfer o ADT_A27: cancel pending admit o ADT_A29: Delete person information o ADT_A45: Move visit information – visit number
> 
> We have added the creation of bundle ID from MSH.10 and reference the same in provenance resource.

Just curious why are we removing these templates?

PR #339, opened 2021-12-06
Add a filter - sliceDataBySegments

yankunhuang-pku:


src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
```diff
@@ -128,5 +128,50 @@ public static bool HasSegments(Hl7v2Data hl7v2Data, string segmentIdContent)
 
             return result;
         }
+
+        public static List<Hl7v2Data> SliceDataBySegments(Hl7v2Data hl7v2Data, List<Hl7v2Segment> segmentList, Hl7v2Segment endSegment = null)
+        {
```
BoyaWu10, line 133: It looks that if `segmentList = new List<Segment>()`, we will return a List<Hl7v2Data> with one element. Maybe consider returning an empty list or throwing exceptions for this case?

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
```diff
@@ -128,5 +128,50 @@ public static bool HasSegments(Hl7v2Data hl7v2Data, string segmentIdContent)
...
+
+                if (startSegmentFound)
+                {
+                    result.Meta.Add(hl7v2Data.Meta[i]);
+                    result.Data.Add(hl7v2Data.Data[i]);
+                    result.Value += valueList[i] + System.Environment.NewLine;
+                }
+            }
+
+            if (result != null)
```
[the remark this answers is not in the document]
yankunhuang-pku: When segmentList is null, the result may be null here.

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
yankunhuang-pku: If segmentList is null, it will return an empty list as expected, and will not return a list with one element.

BoyaWu10: How about an empty segmentList `segmentList = new List<Segment>()`?

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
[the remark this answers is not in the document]
BoyaWu10: When segmentList is null, an exception is thrown instead of returning a result?

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
```diff
@@ -128,5 +128,50 @@ public static bool HasSegments(Hl7v2Data hl7v2Data, string segmentIdContent)
 
             return result;
         }
+
+        public static List<Hl7v2Data> SliceDataBySegments(Hl7v2Data hl7v2Data, List<Hl7v2Segment> segmentList, Hl7v2Segment endSegment = null)
```
BoyaWu10, line 132: Just come up with a strange case: what will happen if `endSegment` is one of the segments in `segmentList`, e.g., the first one, the second one?

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
```diff
@@ -128,5 +128,50 @@ public static bool HasSegments(Hl7v2Data hl7v2Data, string segmentIdContent)
...
+        {
+            var results = new List<Hl7v2Data>();
+            var result = new Hl7v2Data();
+            string[] segmentSeparators = { "\r\n", "\r", "\n" };
+            var valueList = hl7v2Data.Value.Split(segmentSeparators, StringSplitOptions.RemoveEmptyEntries);
```
qiwjin, line 137: I think splitting the input data into segments is a common logic, can we extract it into separated function?

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
qiwjin: need some functional doc

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
```diff
@@ -128,5 +128,50 @@ public static bool HasSegments(Hl7v2Data hl7v2Data, string segmentIdContent)
...
+                        segmentIndex++;
+                    }
+                }
+
+                if (startSegmentFound)
+                {
+                    result.Meta.Add(hl7v2Data.Meta[i]);
+                    result.Data.Add(hl7v2Data.Data[i]);
+                    result.Value += valueList[i] + System.Environment.NewLine;
+                }
```
qiwjin, line 166: maybe use the newline in original test will be better?

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
yankunhuang-pku: If `endsegment` is encountered, it should be stopped even though there are elements not traversed in the `segmentList`. If user don't want to stop it, user should select an appropriate `endsegment`. We can note it in the document.

BoyaWu10: Can we add a unit test about this case?

yankunhuang-pku: Sure~

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
```diff
@@ -128,5 +128,50 @@ public static bool HasSegments(Hl7v2Data hl7v2Data, string segmentIdContent)
...
+        {
+            var results = new List<Hl7v2Data>();
+            var result = new Hl7v2Data();
+            string[] segmentSeparators = { "\r\n", "\r", "\n" };
+            var valueList = hl7v2Data.Value.Split(segmentSeparators, StringSplitOptions.RemoveEmptyEntries);
+            var startSegmentFound = false;
+            var segmentIndex = 0;
+            for (var i = 0; i < hl7v2Data.Meta.Count; ++i)
+            {
+                if (segmentIndex == 0 && segmentList.Count > 0 && ReferenceEquals(hl7v2Data.Data[i], segmentList[segmentIndex]))
```
qiwjin, line 142: just curious, why do we use ReferenceEquals here?

yankunhuang-pku: There may be two identical elements in the `segmentList`. The reason for using it is not only to require that the values of the objects are equal, but also to ensure that the references of the two objects are equal.

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/SegmentFiltersTests.cs
```diff
@@ -116,6 +117,72 @@ public void GivenAnHl7v2Data_WhenHasSegments_CorrectResultShouldBeReturned()
...
+            Assert.Equal(2, orcSlicedData[1].Meta.Count);
+            Assert.Equal(2, orcSlicedData[2].Meta.Count);
+            Assert.Equal(7, orcSlicedData[0].Data.Count);
+            Assert.Equal(2, orcSlicedData[1].Data.Count);
+            Assert.Equal(2, orcSlicedData[2].Data.Count);
+            Assert.Equal("ORC", orcSlicedData[0].Meta[0]);
+            string expectedOrcValue = @"ORC|RE|4422^NIST-AA-IZ-2|13696^NIST-AA-IZ-2|||||||7824^Jackson^Lily^Suzanne^^^^^NIST-PI-1^L^^^PRN||654^Thomas^Wilma^Elizabeth^^^^^NIST-PI-1^L^^^MD|||||NISTEHRFAC^NISTEHRFacility^HL70362|";
+            Assert.Equal(expectedOrcValue, orcSlicedData[0].Data[0].Value);
+
+            // Endsegment precedes the last object in the segmentList
```
BoyaWu10, line 167: Nit: It would be better to add some comments about what result will be returned in this case.

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
BoyaWu10: We can add an utility in engine to do so and also replace the same logics in [Hl7v2DataParser](https://github.com/microsoft/FHIR-Converter/blob/ae582fa986a74def8e1a5736162c62571a315d9e/src/Microsoft.Health.Fhir.Liquid.Converter/Parsers/Hl7v2DataParser.cs#L20).

yankunhuang-pku: Yes, I think so.

src/Microsoft.Health.Fhir.Liquid.Converter/Utilities/Hl7v2DataUtility.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Health.Fhir.Liquid.Converter.Utilities
+{
+    public class Hl7v2DataUtility
```
BoyaWu10, line 7: Can we make this a static class? If so we don't have to new it before using its functions.

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Utilities/Hl7v2DataUtilityTests.cs
```diff
@@ -0,0 +1,31 @@
+﻿using Microsoft.Health.Fhir.Liquid.Converter.Utilities;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Liquid.Converter.UnitTests.Utilities
+{
+    public class Hl7v2DataUtilityTests
```
BoyaWu10, line 6: Can we add some tests when the input message is null or empty?

BoyaWu10 (approved):

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
qiwjin: why not to use "=="?

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
```diff
@@ -128,5 +128,47 @@ public static bool HasSegments(Hl7v2Data hl7v2Data, string segmentIdContent)
...
             return result;
         }
+
+        public static List<Hl7v2Data> SliceDataBySegments(Hl7v2Data hl7v2Data, List<Hl7v2Segment> segmentList, Hl7v2Segment endSegment = null)
+        {
+            var results = new List<Hl7v2Data>();
+            var result = new Hl7v2Data();
+            var startSegmentFound = false;
+            var segmentIndex = 0;
+            for (var i = 0; i < hl7v2Data.Meta.Count; ++i)
```
qiwjin, line 138: i feel the implementation logic is not very clear, maybe it's not the best case to using 'for' here?

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
```diff
@@ -128,5 +128,47 @@ public static bool HasSegments(Hl7v2Data hl7v2Data, string segmentIdContent)
 
             return result;
         }
+
```
qiwjin, line 131: is it necessary to use segment class here?

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/SegmentFiltersTests.cs
```diff
@@ -116,6 +117,72 @@ public void GivenAnHl7v2Data_WhenHasSegments_CorrectResultShouldBeReturned()
             Assert.Throws<NullReferenceException>(() => Filters.HasSegments(new Hl7v2Data(), null));
         }
 
+        [Fact]
```
qiwjin, line 120: also need test for invalid cases

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/SegmentFiltersTests.cs
```diff
@@ -181,12 +195,20 @@ public void GivenAnHl7v2Data_WhenSliceDataBySegments_CorrectResultShouldBeReturn
             // Hl7v2Data and segmentList could not be null
             Assert.Throws<NullReferenceException>(() => Filters.SliceDataBySegments(null, segmentList["ORC"]));
             Assert.Throws<NullReferenceException>(() => Filters.SliceDataBySegments(TestData, null));
+
+            // If the segment in the segmentList / endSegment is not found in the data
```
BoyaWu10, line 199: Let's make the comment complete by adding the returned results.

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/SegmentFiltersTests.cs
```diff
@@ -181,12 +195,20 @@ public void GivenAnHl7v2Data_WhenSliceDataBySegments_CorrectResultShouldBeReturn
...
+            Dictionary<string, List<Hl7v2Segment>> invalidSegmentList = Filters.GetSegmentLists(TestData2, "IN1");
```
BoyaWu10, line 200: Why IN1 is considered invalid? It appears in the source data.

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/SegmentFiltersTests.cs
```diff
@@ -29,7 +29,21 @@ public class SegmentFiltersTests
 ORC|RE||35508^NIST-AA-IZ-2|||||||7824^Jackson^Lily^Suzanne^^^^^NIST-PI-1^L^^^PRN|||||||NISTEHRFAC^NISTEHRFacility^HL70362
 RXA|0|1|20131112||88^influenza, unspecified formulation^CVX|999|||01^Historical Administration^NIP001|||||||||||CP|A";
 
-        private static readonly Hl7v2Data TestData = LoadTestData();
+        private const string TestDataContent2 = @"MSH|^~\&|AccMgr|1|||20050110045504||ADT^A01|599102|P|2.3|||
```
BoyaWu10, line 32: Can we use `TestDataContent` for the test? If not, consider renaming the new data instead of adding index after it.

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
yankunhuang-pku: I think if we use "for" here, we can traverse the data only once to reduce the complexity. Although it may seem obscure, the logic is relatively simple, that is, judge whether to find the starting segment and when to terminate the traversal.

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SegmentFilters.cs
yankunhuang-pku: The advantage of using segment class over string here is that users can also refine the starting point of segmentation, and further customize the segmentList that meets their slicing intention according to their needs, not just through meta of segment.

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/SegmentFiltersTests.cs
yankunhuang-pku: Thanks, added it.

PR #358, opened 2022-03-08
Fix for liquid filter timezone handling

jasonheld-msft:
Currently, all date filters except FormatAsDateTime default to "preserve".  With a default of "local", all functional tests seem to fail depending on what timezone they are run in.  My assumption is that by default the time would not want to be changed when converting to FHIR.


sowu880 (approved):

sowu880:
Thanks a lot for the suggestion. 
After discussion, we decide to change the default time zone handling to "preserve".
And more unit tests are added in PR #363

PR #361, opened 2022-03-14
Add check for large for loops

BoyaWu10:
By offline discussion, we are going to limit the number of the list in the for loop by the `maxIterations` parameter provided by dotliquid. Currently the value is set to 100000.


sowu880 (approved):

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/BaseProcessor.cs
```diff
@@ -18,6 +18,7 @@ namespace Microsoft.Health.Fhir.Liquid.Converter.Processors
     public abstract class BaseProcessor : IFhirConverter
     {
         private readonly ProcessorSettings _settings;
+        private const int _maxIterations = 100000;
```
qiwjin, line 21: can help doc this limitation?

qiwjin (approved):

BoyaWu10:
Hi @ginalee-dotcom, could you please help create a PR to update the document about this change? The maximum iteration of for-loop in liquid templates is now set to 100,000. Previously there's no limitation about this. If the limitation is reached, an exception will be thrown with detailed message. This is to help make sure our engine returns a result in time.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/BaseProcessor.cs
BoyaWu10: Hi @ginalee-dotcom, could you please help create a PR to update the document about this change? The maximum iteration of for-loop in liquid templates is now set to 100,000. Previously there's no limitation about this. If the limitation is reached, an exception will be thrown with detailed message. This is to help make sure our engine returns a result in time.

moria97 (approved):

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/BaseProcessor.cs
```diff
@@ -43,7 +44,7 @@ protected virtual Context CreateContext(ITemplateProvider templateProvider, IDic
                 outerScope: new Hash(),
                 registers: Hash.FromDictionary(new Dictionary<string, object> { { "file_system", templateProvider.GetTemplateFileSystem() } }),
                 errorsOutputMode: ErrorsOutputMode.Rethrow,
-                maxIterations: 0,
+                maxIterations: _maxIterations,
```
ACMoretxj, line 47: Will this configuration effect for the following template?

```liquid
{
    {% for i in (1..10000) -%}
        {% for j in (1..10000) -%}
            "value": [{% for k in (1..10000) -%}{{ k }},{% endfor -%}]
        {% endfor -%}
    {% endfor -%}
}
```

yankunhuang-pku (approved):

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/BaseProcessor.cs
BoyaWu10: No, the limitation only works for single for-loop. An unit test is added to show this.

BoyaWu10: Besides, if the nested for-loops take too much time, it should be killed by timeout setting, which is set by FHIR Server.

ACMoretxj (approved):

PR #376, opened 2022-04-08
Update dockerimage.yml

sanchit11:


.github/workflows/dockerimage.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
     steps:
     - name: 'Checkout Github Action'
-      uses: actions/checkout@master
+      uses: actions/checkout@Handlebars
```
ACMoretxj, line 16: It's might be better to keep align with the case of branch name, i.e. `handlebars`.

irenepjoseph:
@ACMoretxj can this PR be closed with comment?

PR #381, opened 2022-04-15
Update DotLiquid

QuanWanxx:
Update DotLiquid version to 2.2.580, which is most downloads in recent releases from [NuGet org](https://www.nuget.org/packages/DotLiquid/).

Below are some updates summaries refer to DotLiquid [commits](https://github.com/dotliquid/dotliquid/commits/master)

### New Filters
1. where 
2. concat
3. escape_cnce
4. reverse
5. sort_natural
6. upcase_first
7. capitilize
8-12. Md5,Sha1,Sha256,HmacSha1,HmacSha256

### New features
1. Improve template render performance.
2. Support ```enum``` data type, refer [commit](https://github.com/dotliquid/dotliquid/commit/1dc782d31cc13d70f53bcef2ae7dabbe215bbb3e).
3. Support filters with the same method name and different number of parameters from different classes, refer [commit](https://github.com/dotliquid/dotliquid/commit/3efe3e70694bf7a6094a74b33312e7b3c6a36ad0).
4. Support iterating over nested objects, refer [commit](https://github.com/dotliquid/dotliquid/commit/352df87b2f3894788a46d730dd894acac032a7e7).

### Security fixed
Now will throw error/exception for unrecognized filters, see this [issue](https://github.com/dotliquid/dotliquid/issues/413)


data/Templates/Ccda/Resource/_AllergyIntolerance.liquid
```diff
@@ -42,15 +42,13 @@
...
                     "onset": "{{ r.observation.effectiveTime.low.value | format_as_date_time }}", 
                 {% endif -%}
-                {% for templateId in r.observation.templateId -%}
-                    {% if templateId.root == "2.16.840.1.113883.10.20.22.4.8" -%}
-                        {% if r.observation.value.translation -%}
-                            "severity":"{{ r.observation.value.translation.displayName | downcase | get_property: 'ValueSet/AllergySeverity' }}",
-                        {% else -%}
-                            "severity":"{{ r.observation.value.displayName | downcase | get_property: 'ValueSet/AllergySeverity' }}",
-                        {% endif -%}
+                {% if r.observation.templateId | to_json_string | contains: '"2.16.840.1.113883.10.20.22.4.8"' -%}
```
ACMoretxj, line 45: Maybe need to alternate because of the 'if' behavior we discussed today.

QuanWanxx: Will fix all filters in ```if``` tag in another PR, now currently keep the condition check with others

ACMoretxj (approved):

qiwjin:
shall merge this change into dotliquid branch first?

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/SectionFilters.cs
```diff
@@ -96,12 +95,12 @@ public partial class Filters
...
+        private static List<object> GetComponents(IDictionary<string, object> data)
         {
-            var dataComponents = (((data["ClinicalDocument"] as Hash)?
-                ["component"] as Hash)?
-                ["structuredBody"] as Hash)?
-                ["component"];
+            var clinicalDocument = data["ClinicalDocument"] as IDictionary<string, object>;
+            var component = clinicalDocument?.ContainsKey("component") is true ? clinicalDocument["component"] as IDictionary<string, object> : null;
+            var structuredBody = component?.ContainsKey("structuredBody") is true ? component["structuredBody"] as IDictionary<string, object> : null;
+            var dataComponents = structuredBody?.ContainsKey("component") is true ? structuredBody["component"] : null;
```
sowu880, line 103: Could use GetValueOrDefault() ?

sowu880, line 103: Could use GetValueOrDefault() ?

QuanWanxx: Here ```IDictionary``` only have ```bool TryGetValue(string key, out value)``` method, the ```GetValueOrDefault()``` belongs to ```Dictionary``` which may not the exact instance here.

sowu880: Do we need to transform it to IDictionary? Since element data["ClinicalDocument"] is a Dictionary.

sowu880: 'IDictionary' is better here.

QuanWanxx: From offline discussions, still keep 'IDictionary' here in case we internally use instances that are not dictionaries in the future.

qiwjin: why not use TryGetValue here?

QuanWanxx: The ```bool TryGetValue(TKey key, out TValue value)``` still need us to define some intermediate variable in ```out value``` parameter, I'll take define an IDictionary extension so the code here can be simplified.

sowu880: For ccda input data we parse it to `Dictionary` object https://github.com/microsoft/FHIR-Converter/blob/027f35519d72c5b08fd24c45b8c029958919716a/src/Microsoft.Health.Fhir.Liquid.Converter/DictionaryJsonConverter.cs#L85
 but `Idictionary` outside, do we need to make them consistent? https://github.com/microsoft/FHIR-Converter/blob/027f35519d72c5b08fd24c45b8c029958919716a/src/Microsoft.Health.Fhir.Liquid.Converter/Parsers/CcdaDataParser.cs#L46

If they are ccda specific filters, the input will always be `Dictionary` for now.

QuanWanxx: Here we assume the data structure always be like ```data["ClinicalDocument"]["component"]["structuredBody"]["component"]```, which indicates the input must be **ccda** format.

I think the using ```Dictionary``` internally can be enough for now, if we update this ```SectionFilter``` to take more input format in the future, we can then use ```IDictionary``` instead.

QuanWanxx: Have update to ```as Dictionary``` and use ```GetValueOrDefault```, same with other code in this filter.

sowu880 (approved):

qiwjin (approved):

Issue #401, opened 2022-06-15
format_as_date_time - conversion issue


aunabidi:
I am trying to convert HL7 messages to FHIR and run into a datetime conversion error. The datetime "20140630120030.123**4**-0500" errors out with the message below but it works fine when I remove the 4 in bold from the example above and use "20140630120030.123-0500"

Error:
"Failed to convert the input data. Reason: The format of Date or DateTime is invalid: 20140630120030.1234-0500."

According to the definition of format_as_date_time it should accept the 4th digit in the second: YYYY[MM[DD[HH[MM[SS[.S[S[S[S]]]]]]]]][+/-ZZZZ] but it throws an error.

Any help would be appreciated. Thanks!

yankunhuang-pku:
Could you help to investigate this issue? @sowu880

aunabidi:
Hi @sowu880. Any luck with this issue?

Thanks!

aunabidi:
@sowu880 @yankunhuang-pku I see that the issue was fixed, when will this fix be available in Azure FHIR API?

PR #409, opened 2022-06-30
Fix more than 3 digits for milliseconds in datetime

sowu880:
Fix issue #401


src/Microsoft.Health.Fhir.Liquid.Converter/Models/PartialDateTime.cs
```diff
@@ -37,7 +37,8 @@ public PartialDateTime(string input, DateTimeType type = DateTimeType.Fhir)
             int hour = groups["hour"].Success ? int.Parse(groups["hour"].Value) : 0;
             int minute = groups["minute"].Success ? int.Parse(groups["minute"].Value) : 0;
             int second = groups["second"].Success ? int.Parse(groups["second"].Value) : 0;
-            int millisecond = groups["millisecond"].Success ? int.Parse(groups["millisecond"].Value) : 0;
+            int millisecond = groups["millisecond"].Success ? int.Parse(groups["millisecond"].Value.Length > 3 ? groups["millisecond"].Value.Substring(0, 3) : groups["millisecond"].Value) : 0;
```
yankunhuang-pku, line 40: The 4th digit is invalid?

yankunhuang-pku (approved):

src/Microsoft.Health.Fhir.Liquid.Converter/Models/PartialDateTime.cs
sowu880: 4th is invalid when new a datetime object in .net.
So we both save string part of millisecond and 3 digits as int value here.

PR #440, opened 2022-11-22
Change Mediatype to v2

sowu880:
1. Update oras version from 0.12.0 to 0.16.0
2. Change media type of image from `application/vnd.oci.image.manifest.v1+json` to `application/vnd.docker.distribution.manifest.v2+json`


sowu880:
Abandon this changes.

PR #460, opened 2023-03-17
fixed #458

### LINKED ISSUE #458: is_nan filter issue
The documentation calls out 'is_nan' as a filter. The example given is:

image

After doing some testing with this it seems to actually use the filter in liquid you would have to write the filter as 'in_na_n'. Using 'is_nan' results in the variable you are checking being returned as the result since it doesn't seem to be recognized as a valid filter.

I'm not sure what is converting the dotnet method to a filter, but it seems to key off of casing. Since the method is named 'IsNaN' I think that's resulting in the liquid filter being named 'is_na_n'. If the method were named 'IsNan' I think the filter would then match the documentation.

image

Also the result of the filter seems to be reversed, things that are numbers are returning 'true' and things that are not numbers are returning 'false'. Below is the code I used to test, and the output.

Liquid code:
image

Json output:
image

JSpinks93:
Addressed issues in #458

Adjusted name of IsNan MathFilter so that it can be used as 'is_nan' in liquid templates.

Fixed logic for IsNan filter as it was returning the opposite of what it was supposed to.


JSpinks93:
@microsoft-github-policy-service agree company="3Cloud"

PR #485, opened 2023-08-07
Support reading convert templates from a storage container

wi-y:
Adds classes and interfaces to support reading convert templates from a storage container

### DIFF
--- a/src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
+++ b/src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
@@ -33,7 +33,6 @@ namespace Microsoft.Health.Fhir.TemplateManagement.ArtifactProviders
 
         private readonly AsyncRetryPolicy _downloadRetryPolicy;
 
-
         public BlobTemplateProvider(BlobContainerClient blobContainerClient, IMemoryCache templateCache, TemplateCollectionConfiguration templateConfiguration)
         {
             _blobContainerClient = blobContainerClient;
@@ -93,7 +92,7 @@ namespace Microsoft.Health.Fhir.TemplateManagement.ArtifactProviders
         {
             var blobClient = blobContainerClient.GetBlobClient(blobName);
 
-            PolicyResult<Response<BlobDownloadResult>> policyResponse = await _downloadRetryPolicy.ExecuteAndCaptureAsync(async() => await blobClient.DownloadContentAsync(ct));
+            PolicyResult<Response<BlobDownloadResult>> policyResponse = await _downloadRetryPolicy.ExecuteAndCaptureAsync(async () => await blobClient.DownloadContentAsync(ct));
 
             if (policyResponse.FinalException == null)
             {
--- a/src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
+++ b/src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
@@ -13,6 +14,8 @@ using DotLiquid;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Health.Fhir.Liquid.Converter.Utilities;
 using Microsoft.Health.Fhir.TemplateManagement.Models;
+using Polly;
+using Polly.Retry;
 
 namespace Microsoft.Health.Fhir.TemplateManagement.ArtifactProviders
 {
@@ -28,19 +31,25 @@ namespace Microsoft.Health.Fhir.TemplateManagement.ArtifactProviders
 
         private readonly int _segmentSize = 100;
 
+        private readonly AsyncRetryPolicy _downloadRetryPolicy;
+
+
         public BlobTemplateProvider(BlobContainerClient blobContainerClient, IMemoryCache templateCache, TemplateCollectionConfiguration templateConfiguration)
         {
             _blobContainerClient = blobContainerClient;
             _templateCache = templateCache;
             _templateCollectionConfiguration = templateConfiguration;
+            _downloadRetryPolicy = Policy
+                    .Handle<Exception>()
+                    .WaitAndRetryAsync(3, retryAttempt => TimeSpan.FromMilliseconds(10));
         }
 
         public async Task<List<Dictionary<string, Template>>> GetTemplateCollectionAsync(CancellationToken cancellationToken = default)
         {
             // read templates from cache if available
-            if (_templateCache.TryGetValue(_blobTemplateCacheKey, out var templateCache))
+            if (_templateCache.TryGetValue(_blobTemplateCacheKey, out List<Dictionary<string, Template>> templateCache))
             {
-                return (List<Dictionary<string, Template>>)templateCache;
+                return templateCache;
             }
 
             var templateNames = await ListBlobsFlatListing(_blobContainerClient, _segmentSize, cancellationToken);
@@ -49,7 +58,7 @@ namespace Microsoft.Health.Fhir.TemplateManagement.ArtifactProviders
 
             foreach (var templateName in templateNames)
             {
-                var template = await DownloadBlobToStringAsync(_blobContainerClient, templateName);
+                var template = await DownloadBlobToStringAsync(_blobContainerClient, templateName, cancellationToken);
                 templates.Add(templateName, template);
             }
 
@@ -66,7 +75,7 @@ namespace Microsoft.Health.Fhir.TemplateManagement.ArtifactProviders
         {
             var blobs = new List<string>();
 
-            var resultSegment = blobContainerClient.GetBlobsAsync()
+            var resultSegment = blobContainerClient.GetBlobsAsync(default, default, null, ct)
                 .AsPages(default, segmentSize);
 
             await foreach (Page<BlobItem> blobPage in resultSegment)
@@ -80,11 +89,20 @@ namespace Microsoft.Health.Fhir.TemplateManagement.ArtifactProviders
             return blobs;
         }
 
-        public static async Task<string> DownloadBlobToStringAsync(BlobContainerClient blobContainerClient, string blobName)
+        public async Task<string> DownloadBlobToStringAsync(BlobContainerClient blobContainerClient, string blobName, CancellationToken ct)
         {
             var blobClient = blobContainerClient.GetBlobClient(blobName);
-            BlobDownloadResult downloadResult = await blobClient.DownloadContentAsync();
-            return downloadResult.Content.ToString();
+
+            PolicyResult<Response<BlobDownloadResult>> policyResponse = await _downloadRetryPolicy.ExecuteAndCaptureAsync(async() => await blobClient.DownloadContentAsync(ct));
+
+            if (policyResponse.FinalException == null)
+            {
+                return policyResponse.Result.Value.Content.ToString();
+            }
+            else
+            {
+                throw policyResponse.FinalException;
+            }
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
+++ b/src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
@@ -1,0 +1,99 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure;
+using Azure.Storage.Blobs;
+using Azure.Storage.Blobs.Models;
+using DotLiquid;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Health.Fhir.Liquid.Converter.Utilities;
+
+namespace Microsoft.Health.Fhir.TemplateManagement.ArtifactProviders
+{
+    public class BlobTemplateProvider : IConvertDataTemplateProvider
+    {
+        private BlobContainerClient _blobContainerClient;
+
+        private readonly IMemoryCache _templateCache;
+
+        private readonly string _blobTemplateCacheKey = "cached-blob-templates";
+
+        private readonly int _segmentSize = 100;
+
+        public BlobTemplateProvider(BlobContainerClient blobContainerClient, IMemoryCache templateCache)
+        {
+            _blobContainerClient = blobContainerClient;
+            _templateCache = templateCache;
+        }
+
+        public async Task<List<Dictionary<string, Template>>> GetTemplateCollectionAsync(CancellationToken cancellationToken = default)
+        {
+            // read templates from cache if available
+            if (_templateCache.TryGetValue(_blobTemplateCacheKey, out var templateCache))
+            {
+                return (List<Dictionary<string, Template>>)templateCache;
+            }
+
+            var templateLookup = new Dictionary<string, Template>();
+            var templateNames = await ListBlobsFlatListing(_blobContainerClient, _segmentSize);
+
+            foreach (var templateName in templateNames)
+            {
+                var template = await DownloadBlobToStringAsync(_blobContainerClient, templateName);
+                var renamedTemplate = TrimBlobName(templateName);
+                Template liquidTemplate = TemplateUtility.ParseTemplate(renamedTemplate, template);
+                templateLookup.Add(renamedTemplate, liquidTemplate);
+            }
+
+            var templates = new List<Dictionary<string, Template>>
+            {
+                templateLookup,
+            };
+
+            _templateCache.Set(_blobTemplateCacheKey, templates);
+
+            return templates;
+        }
+
+        private static async Task<List<string>> ListBlobsFlatListing(BlobContainerClient blobContainerClient, int? segmentSize)
+        {
+            var blobs = new List<string>();
+
+            var resultSegment = blobContainerClient.GetBlobsAsync()
+                .AsPages(default, segmentSize);
+
+            await foreach (Page<BlobItem> blobPage in resultSegment)
+            {
+                foreach (BlobItem blobItem in blobPage.Values)
+                {
+                    blobs.Add(blobItem.Name);
+                }
+            }
+
+            return blobs;
+        }
+
+        public static async Task<string> DownloadBlobToStringAsync(BlobContainerClient blobContainerClient, string blobName)
+        {
+            var blobClient = blobContainerClient.GetBlobClient(blobName);
+            BlobDownloadResult downloadResult = await blobClient.DownloadContentAsync();
+            return downloadResult.Content.ToString();
+        }
+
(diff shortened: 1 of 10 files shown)


src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+    {
+        private BlobContainerClient _blobContainerClient;
+
+        private readonly IMemoryCache _templateCache;
+
+        private readonly string _blobTemplateCacheKey = "cached-blob-templates";
+
+        private readonly int _segmentSize = 100;
+
+        public BlobTemplateProvider(BlobContainerClient blobContainerClient, IMemoryCache templateCache)
```
dustinburson, line 28: It looks like BlobContainerClient should be mockable, https://devblogs.microsoft.com/azure-sdk/unit-testing-and-mocking/

If so we should have unit tests for this class to verify the logic.

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+                Template liquidTemplate = TemplateUtility.ParseTemplate(renamedTemplate, template);
+                templateLookup.Add(renamedTemplate, liquidTemplate);
+            }
+
+            var templates = new List<Dictionary<string, Template>>
+            {
+                templateLookup,
+            };
+
+            _templateCache.Set(_blobTemplateCacheKey, templates);
```
dustinburson, line 58: What is the policy for expiring entries in the cache?  How does that compare with the ACR policy?

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/IConvertDataTemplateProvider.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using DotLiquid;
+
+namespace Microsoft.Health.Fhir.TemplateManagement.ArtifactProviders
+{
+    public interface IConvertDataTemplateProvider
+    {
+        public Task<List<Dictionary<string, Template>>> GetTemplateCollectionAsync(CancellationToken cancellationToken);
```
pallar-ms, line 10: Could you clarify the intended returned object? 
It seems like the dictionary is (templateName, templateContents), so the list is to capture multiple sets of template collections?

src/Microsoft.Health.Fhir.TemplateManagement/Factory/StorageTemplateProviderFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+            _blobContainerClient = blobContainerClient;
+        }
+
+        public IConvertDataTemplateProvider GetContainerRegistryTemplateProvider()
+        {
+            throw new NotImplementedException();
+        }
+
+        public IConvertDataTemplateProvider GetDefaultTemplateProvider()
+        {
```
pallar-ms, line 33: Just my 2c, since this factory class is named specific to StorageTemplate, on first glance, even though I know it is implementing ITemplateProviderFactory, it feels odd to see GetContainerRegistryTemplateProvider() and GetDefaultTemplateProvider() in here.
I think making this factory class more generic in interpretation then would help.
But if this is the pattern/practice for factory providers, then feel free to ignore.

src/Microsoft.Health.Fhir.TemplateManagement/Factory/StorageTemplateProviderFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+            throw new NotImplementedException();
+        }
+
+        public IConvertDataTemplateProvider GetStorageContainerTemplateProvider()
+        {
+            if (_memoryCache.TryGetValue(_templateProviderCachePrefix + _blobContainerClient.Name, out var templateProviderCache) == true)
```
pallar-ms, line 39: nit: Can save this in var and use below as well. Just so we don't miss if one gets updated and not the other.

pallar-ms, line 39: nit: can remove '== true'.

src/Microsoft.Health.Fhir.TemplateManagement/Factory/StorageTemplateProviderFactory.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                return (IConvertDataTemplateProvider)templateProviderCache;
+            }
+
+            var templateProvider = new BlobTemplateProvider(_blobContainerClient, _memoryCache);
+            _memoryCache.CreateEntry(_templateProviderCachePrefix + _blobContainerClient.Name).Value = new BlobTemplateProvider(_blobContainerClient, _memoryCache);
```
pallar-ms, line 45: Should this instead be the above created templateProvider object?

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
```diff
@@ -0,0 +1,99 @@
...
+            BlobDownloadResult downloadResult = await blobClient.DownloadContentAsync();
+            return downloadResult.Content.ToString();
+        }
+
+        // TODO: eventually want to remove this and create a common utility for ACR and storage templates
+        private string TrimBlobName(string templateName)
+        {
+            return templateName
+                .Replace("Hl7v2/_", string.Empty)
+                .Replace("Hl7v2/", string.Empty)
```
pallar-ms, line 93: Should this include the folder prefix for other formats too like ccda, json, etc.?

src/Microsoft.Health.Fhir.TemplateManagement/Factory/StorageTemplateProviderFactory.cs
dustinburson: Looking at the existing code, the template provider contract is specific to the ACR implementation.  I think a larger refactor is order to align a common interface for consumers regardless of the source.

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
wi-y: Right now for this PR the cache for storage templates won't expire. I can pass some cache options in. The ACR cache logic is pretty confusing and not sure we want to replicate it. There are long and short expiry times of 60 minutes and 20 minutes here: https://github.com/microsoft/FHIR-Converter/blob/baf921359046571421d26f52d8360d8b34045d76/src/Microsoft.Health.Fhir.TemplateManagement/Models/TemplateCollectionConfiguration.cs#L10

But also the template provider itself seems to only be cached for 5 minutes (so I think 5 minutes is the real value since the template provider gets re-initialized every 5 minutes and therefore anything in the 20 minute cache or 60 minute cache seems like it would be lost): https://github.com/microsoft/fhir-server/blob/536a1021d32226645f8e8338345ecdc90afea198/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ContainerRegistryTemplateProvider.cs#L98C29-L98C51

I was thinking we could have that data plane operation we proposed for template settings locations(s) and template settings in general could be used to invalidate the cache, or set the cache time. So at the very least I should probably pass some cache options in.

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
wi-y: Thanks, I've added unit tests

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
wi-y: added options for caching

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/IConvertDataTemplateProvider.cs
wi-y: I think the list was intended to capture a base template collection and a user defined template collection. It may be useful in the future if we figure out a good way to overwrite the base settings with the users defined settings. For now we'll just assume that the user has uploaded all of the necessary templates to the storage account so there will only be one entry in the list.

The FHIR service defined a similar interface [here](https://github.com/microsoft/fhir-server/blob/10e30c9a8e977d674596d1dad731338a062029e2/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/IConvertDataTemplateProvider.cs#L14) which is useful but I didn't like that it accepted a ConvertData request as a parameter.

src/Microsoft.Health.Fhir.TemplateManagement/Factory/StorageTemplateProviderFactory.cs
wi-y: Thanks for the feedback, agreed. I like the interface ITemplateProviderFactory just having a single method for GetTemplateProvider instead of the interface having methods for each template source. Then the factory can be injected based on options specified. There is still some cleanup required for some of the ACR template providers to not have ACR specific methods but they should be able to be switched to a generic ITemplateProviderFactory / GetTemplateProvider() method

src/Microsoft.Health.Fhir.TemplateManagement/Factory/StorageTemplateProviderFactory.cs
wi-y: thanks, good catch. updated

wi-y: updated

src/Microsoft.Health.Fhir.TemplateManagement/Factory/StorageTemplateProviderFactory.cs
wi-y: yes, good catch, that would make more sense. updated

wi-y: yes, good catch, that would make more sense. updated

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
wi-y: Yeah it should. I ended up switching to TemplateUtility.ParseTemplates which handles the filename parsing. I had been incorrectly using TemplateUtility.ParseTemplate previously which did not handle the filename parsing so this helper method was written but now that we've switched to TemplateUtility.ParseTemplates the method is no longer required so it has been removed.

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
dustinburson: Thanks Will, this could be something we incorporate into the new API design as well. Maybe we allow the user to bypass cache based on a header or param when specified.

Thank you for breakdown of the ACR policy.  What you shared mirrors my understanding of the code which was slightly confusing to me because one of the reasons shared for using ACR was to have a single hash for comparison to determine the cache should be invalidated.  So I was expecting we would be at least validating that the hash hasn't changed at the source with each request but that doesn't appear to be the case.

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            _blobContainerClient = blobContainerClient;
+            _templateCache = templateCache;
+            _templateCollectionConfiguration = templateConfiguration;
+        }
+
+        public async Task<List<Dictionary<string, Template>>> GetTemplateCollectionAsync(CancellationToken cancellationToken = default)
+        {
+            // read templates from cache if available
+            if (_templateCache.TryGetValue(_blobTemplateCacheKey, out var templateCache))
```
pallar-ms, line 41: nit: can we use the explicit type (List<Dictionary<string, Template>>) in the out param and skip the cast?

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+            _templateCache.Set(_blobTemplateCacheKey, templateCollection, _templateCollectionConfiguration.ShortCacheTimeSpan);
+
+            return templateCollection;
+        }
+
+        private static async Task<List<string>> ListBlobsFlatListing(BlobContainerClient blobContainerClient, int? segmentSize, CancellationToken ct)
+        {
+            var blobs = new List<string>();
+
+            var resultSegment = blobContainerClient.GetBlobsAsync()
```
pallar-ms, line 69: We should consider passing the cancellation token here.

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+                }
+            }
+
+            return blobs;
+        }
+
+        public static async Task<string> DownloadBlobToStringAsync(BlobContainerClient blobContainerClient, string blobName)
+        {
+            var blobClient = blobContainerClient.GetBlobClient(blobName);
+            BlobDownloadResult downloadResult = await blobClient.DownloadContentAsync();
```
pallar-ms, line 86: We should consider having this be a cancellable operation.

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+                foreach (BlobItem blobItem in blobPage.Values)
+                {
+                    blobs.Add(blobItem.Name);
+                }
+            }
+
+            return blobs;
+        }
+
+        public static async Task<string> DownloadBlobToStringAsync(BlobContainerClient blobContainerClient, string blobName)
```
pallar-ms, line 83: What should our error handing here be? If we encounter some exception when downloading a specific template, I think we should continue to attempt downloading others with a retry policy (with backoff) on this operation.

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+            _templateCache = templateCache;
+            _templateCollectionConfiguration = templateConfiguration;
+        }
+
+        public async Task<List<Dictionary<string, Template>>> GetTemplateCollectionAsync(CancellationToken cancellationToken = default)
+        {
+            // read templates from cache if available
+            if (_templateCache.TryGetValue(_blobTemplateCacheKey, out var templateCache))
+            {
+                return (List<Dictionary<string, Template>>)templateCache;
```
dustinburson, line 43: Not sure about a the public contract returning the entire cache, especially since the we would want to use the IMemoryCache to enforce the lifecycle of the cached item.  Once the dictionary is returned we don't control what is done with the object and it can potentially out live the cache policy we set.  I think a better strategy would be hide that detail from the caller and only return the exact template needed and like the internal logic determine the result.

Another question, what is the cache key in this instance?  Is it the storage location or the individual template?  I am trying to reconcile why inside the cache is a dictionary of templates rather than a single template.

All that said, I realize some of this is due to the existing implementation.  I am fine deferring these changes until we look a unifying the template logic across sources.

dustinburson (dismissed):

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
dustinburson: We should be careful of waiting too long with the backoff policy to avoid unnecessary delays in call times but as long as the timeframe is in the milliseconds range I agree.

src/Microsoft.Health.Fhir.TemplateManagement/Factory/StorageTemplateProviderFactory.cs
wi-y: I see, updated now

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
wi-y: I added some retries for throwing an exception. I think we'll throw for now but I can see us adding more sophisticated logic in the future

src/Microsoft.Health.Fhir.TemplateManagement/Factory/StorageTemplateProviderFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+        public StorageTemplateProviderFactory(BlobContainerClient blobContainerClient, IMemoryCache memoryCache, TemplateCollectionConfiguration templateCollectionConfiguration)
+        {
+            _blobContainerClient = blobContainerClient;
+            _memoryCache = memoryCache;
+            _templateCollectionConfiguration = templateCollectionConfiguration;
+        }
+
+        public IConvertDataTemplateProvider GetTemplateProvider()
+        {
+            var cacheKey = _templateProviderCachePrefix + _blobContainerClient.Name;
```
pallar-ms, line 32: nit: Seems like 'cacheKey' can also be an instance variable set upon construction.

src/Microsoft.Health.Fhir.TemplateManagement/Factory/StorageTemplateProviderFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+        private IMemoryCache _memoryCache;
+
+        private TemplateCollectionConfiguration _templateCollectionConfiguration;
+
+        private string _templateProviderCachePrefix = "storage-template-provider-";
+
+        public StorageTemplateProviderFactory(BlobContainerClient blobContainerClient, IMemoryCache memoryCache, TemplateCollectionConfiguration templateCollectionConfiguration)
+        {
+            _blobContainerClient = blobContainerClient;
+            _memoryCache = memoryCache;
```
pallar-ms, line 26: nit: EnsureArg.IsNotNull() for these.

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
```diff
@@ -0,0 +1,108 @@
...
+            _templateCache.Set(_blobTemplateCacheKey, templateCollection, _templateCollectionConfiguration.ShortCacheTimeSpan);
+
+            return templateCollection;
+        }
+
+        private static async Task<List<string>> ListBlobsFlatListing(BlobContainerClient blobContainerClient, int? segmentSize, CancellationToken ct)
+        {
+            var blobs = new List<string>();
+
+            var resultSegment = blobContainerClient.GetBlobsAsync(default, default, null, ct)
```
pallar-ms, line 78: nit: Suggestion - adding param names for when we aren't passing variables will help indicate what is being passed.

pallar-ms (approved):

pallar-ms (approved):

PR #489, opened 2023-08-17
Change default to not log the input message in any post-processing exceptions

wi-y:
If any exceptions occur during post processing, the current state is that the input message is included by default in the exception. Since the input message can contain sensitive information, we should instead not log this information by default, even though it is useful for troubleshooting. This PR changes the default behavior such that the input message is no longer logged with the post-processing exception.

### DIFF
--- a/src/Microsoft.Health.Fhir.Liquid.Converter/OutputProcessors/PostProcessor.cs
+++ b/src/Microsoft.Health.Fhir.Liquid.Converter/OutputProcessors/PostProcessor.cs
@@ -18,13 +18,14 @@ namespace Microsoft.Health.Fhir.Liquid.Converter.OutputProcessors
 {
     public static class PostProcessor
     {
-        public static JObject Process(string input)
+        public static JObject Process(string input, ProcessorSettings processorSettings)
         {
-            var jsonObject = ParseJson(input);
-            return MergeJson(jsonObject);
+            bool includeInputInException = processorSettings.IncludeInputInException;
+            var jsonObject = ParseJson(input, includeInputInException);
+            return MergeJson(jsonObject, includeInputInException);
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Fhir.Liquid.Converter/OutputProcessors/PostProcessor.cs
```diff
@@ -18,13 +18,14 @@ namespace Microsoft.Health.Fhir.Liquid.Converter.OutputProcessors
 {
     public static class PostProcessor
     {
-        public static JObject Process(string input)
+        public static JObject Process(string input, ProcessorSettings processorSettings)
```
wi-y, line 21: An alternative solution would be to throw the PostprocessException as-is

and then further downstream in the fhir-server code we could intercept the exception, return the details to the user, but NOT log some of the inner details which could contain sensitive information

We could potentially intercept the exception here:
https://github.com/microsoft/fhir-server/blob/00facae70124c40d5f05fcd26fd504f98cbaced5/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ConvertDataEngine.cs#L90C43-L90C59

It is complicated though because I believe all exceptions are currently logged so it would probably involve changing the exception details. It is something I am currently investigating

wi-y: Did some more research on the fhir-server repo. I think we can suppress the PostprocessException from the logs there instead, but still return the relevent postprocess errors the user in the HTTP response. Created a PR in the fhir-server repo: https://github.com/microsoft/fhir-server/pull/3486/files and will abandon this PR

PR #490, opened 2023-08-21
Add telemetry metrics for FHIR converter.

ms-teli:
- Add dependency from iomt-fhir. The dependencies should be moved to the health-shared-components at a later point.
- Add metrics for Converter.
- Update Unit Tests to report the metrics.


nuget.config
```diff
@@ -8,6 +8,7 @@
 		<clear />
 		<add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
 		<add key="Microsoft Health OSS" value="https://microsofthealthoss.pkgs.visualstudio.com/FhirServer/_packaging/Public/nuget/v3/index.json" />
+		<add key="Microsoft MedTech OSS" value="https://microsofthealthoss.pkgs.visualstudio.com/MedTech/_packaging/MedTechOSS/nuget/v3/index.json" />
```
pallar-ms, line 11: I'm guessing this is needed to pull in Microsoft.Health.Common.
Out of scope for this PR, but I think we should move common components (including telemetry) to a different feed. It is not intuitive as to why we need MedTech components in here otherwise. 
But that is a larger refactor, just wanted to call out.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests.csproj
```diff
@@ -15,6 +15,7 @@
   <ItemGroup>
     <PackageReference Include="Hl7.Fhir.R4" Version="2.0.3" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.7.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" NoWarn="NU1605" />
```
pallar-ms, line 18: Disable warning intended?

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/CcdaProcessor.cs
```diff
@@ -4,26 +4,34 @@
...
         {
         }
 
-        public override string Convert(string data, string rootTemplate, ITemplateProvider templateProvider, TraceInfo traceInfo = null)
+        protected override string InternalConvert(string data, string rootTemplate, ITemplateProvider templateProvider, TraceInfo traceInfo = null)
         {
-            var ccdaData = _parser.Parse(data);
-            return Convert(ccdaData, rootTemplate, templateProvider, traceInfo);
+            object ccdaData;
+            using (ITimed templateRetrivalTime = TelemetryLogger.TrackDuration(ConverterMetrics.InputDeserializationDuration))
```
pallar-ms, line 29: nit: rename `templateRetrivalTime` to reflect the metric InputDeserializationDuration. 
Same for below classes.

src/Microsoft.Health.Fhir.Liquid.Converter/Telemetry/ConverterMetrics.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft.Health.Fhir.Liquid.Converter.Telemetry
+{
+    public sealed class ConverterMetrics
+    {
+        public static Metric TotalDuration { get => new Metric(nameof(TotalDuration), new Dictionary<string, object> { }); }
+
+        public static Metric InputDeserializationDuration { get => new Metric(nameof(InputDeserializationDuration), new Dictionary<string, object> { }); }
+
+        public static Metric TemplateRetrivalDuration { get => new Metric(nameof(TemplateRetrivalDuration), new Dictionary<string, object> { }); }
```
pallar-ms, line 17: nit: typo. Instead ` TemplateRetrievalDuration'.

src/Microsoft.Health.Fhir.Liquid.Converter/Telemetry/ConverterMetrics.cs
```diff
@@ -0,0 +1,23 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using Metric = Microsoft.Health.Common.Telemetry.Metric;
+
+namespace Microsoft.Health.Fhir.Liquid.Converter.Telemetry
+{
+    public sealed class ConverterMetrics
+    {
+        public static Metric TotalDuration { get => new Metric(nameof(TotalDuration), new Dictionary<string, object> { }); }
```
pallar-ms, line 13: nit: Can we specify the param name here to indicate what the dictionary is being passed for.

pallar-ms: I'm guessing dimensions?

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/BaseProcessor.cs
```diff
@@ -12,28 +12,52 @@
...
+                result = InternalConvert(data, rootTemplate, templateProvider, traceInfo);
+            }
+
+            return result;
+        }
+
+        public string Convert(string data, string rootTemplate, ITemplateProvider templateProvider, TraceInfo traceInfo = null)
+        {
+            string result;
+            using (ITimed totalConversionTime = TelemetryLogger.TrackDuration(ConverterMetrics.TotalDuration))
```
pallar-ms, line 52: What is the best way to associate these metrics to the operation status, i.e., pass/fail. Should we consider adding dimensions to also capture whether the overall conversion succeeded or failed.
If say the conversion fails early (in case one of the inner steps throws early), the duration could be short and give incorrect impressions on the performance.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirProcessor.cs
```diff
@@ -4,24 +4,25 @@
...
-        public override string Convert(string data, string rootTemplate, ITemplateProvider templateProvider, TraceInfo traceInfo = null)
+        protected override string InternalConvert(string data, string rootTemplate, ITemplateProvider templateProvider, TraceInfo traceInfo = null)
         {
             // TO DO For FHIR Specific Logic
             // 1. Add FHIR related checks for pre-processor and post-processor.
             // 2. Add log for version conversion.
             // 3. Add logic to process "extension" and "contained" fields.
 
-            return base.Convert(data, rootTemplate, templateProvider, traceInfo);
+            return base.InternalConvert(data, rootTemplate, templateProvider, traceInfo);
```
pallar-ms, line 25: Should be `base.InternalConvertFromObject` instead.

pallar-ms (changes requested):

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests.csproj
ms-teli: Yes. It was intentionally disabled. I kept the current version because the latest version of Newtonsoft has some breaking change.

nuget.config
ms-teli: I agree. and I also agree that this should be handled in another scope of work.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/CcdaProcessor.cs
ms-teli: Right, thanks for pointing out.

src/Microsoft.Health.Fhir.Liquid.Converter/Telemetry/ConverterMetrics.cs
ms-teli: thanks for pointing out!

src/Microsoft.Health.Fhir.Liquid.Converter/Telemetry/ConverterMetrics.cs
ms-teli: That makes sense.  Updated.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirProcessor.cs
ms-teli: Actually, this for now should be "base.Convert" defined in JsonProcessor. It's now just reusing logic of Json transformation.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/BaseProcessor.cs
ms-teli: I agree with your suggestion. So far, this change is to add basic touch points to evaluate the duration of transformation and its each steps. I also believe that we can extend this to have more dimensions and create metrics on our backend to monitor. A precondition of doing that would be a general overview of metrics/alarms strategy. We also need to consider the implication and recommendations for external customers when we update the OSS. I would think another task to wrap up that scope. :)

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirProcessor.cs
pallar-ms: Oh totally missed the fact that it is inheriting from JsonProcessor and not BaseProcessor.

pallar-ms (approved):

PR #495, opened 2023-10-26
Download blob templates in parallel for better perf

wi-y:
Updates the blob template download code to download templates in parallel.


src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateProvider.cs
```diff
@@ -54,14 +57,32 @@ public BlobTemplateProvider(BlobContainerClient blobContainerClient, IMemoryCach
...
 
-            var templates = new Dictionary<string, string>();
+            var templates = new ConcurrentDictionary<string, string>();
+
+            // download blobs in parallel
+            var semaphore = new SemaphoreSlim(_maxParallelism);
+            var downloadTasks = new List<Task>();
 
             foreach (var templateName in templateNames)
             {
```
pallar-ms, line 67: To clarify, this is downloading the root template folders in parallel right? i.e., Hl7v2/, CCDA/, etc. are in parallel?
If so, I think it is also common for customers to only have one of these template folders, say just Hl7v2, in which case we would still take this perf hit.
Can we modify to parallelize based on blob count instead?

wi-y: The intent is to download the files in the container in parallel, so if a user only has an Hl7v2 folder, then the files in the folder will be still be downloaded in parallel. Previously, each file in the container was downloaded in order, and we did not attempt to download the next file until the current file was downloaded. ListBlobsFlatListing I believe is what gets every blob in the container and then later we attempt to download 50 at a time until everything is downloaded

pallar-ms (approved):

PR #497, opened 2023-11-09
Add Appointment reference in Encounter & Fill subject in Appointment

umar-gh:


umar-gh:
Submitting another cherrypicked PR

PR #499, opened 2023-11-13
add appointment reference in encounter

umar-gh:


umar-gh:
@microsoft-github-policy-service agree

PR #507, opened 2023-12-05
implement classes for FHIR to HL7v2 conversion

wi-y:


src/Microsoft.Health.Fhir.Liquid.Converter/Models/Hl7v2/MinimalHl7v2Message.cs
```diff
@@ -0,0 +1,26 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+
+namespace Microsoft.Health.Fhir.Liquid.Converter.Models.Hl7v2
+{
+    public class MinimalHl7v2Message
```
dustinburson, line 10: A couple questions here:
* What is the reasoning for the new classes over the the existing HL7 models?  The current ones are setup for use via Liquid, does that implementation block usage for our scenarios?  I see you also added a name property to the segment class.  If that would be useful to the other the ingest models it we be nice if we could reuse.
*  From a naming standpoint we might want another namespace or different name to better indicate the usage.  For example these models wouldn't be used for ingest since they aren't compatible with .NET Liquid as is.  Perhaps a name/namespace that indicates they are used for output only if we can combined models.

data/SampleData/Json/bundle.json
```diff
@@ -0,0 +1,302 @@
+{
+  "resourceType": "Bundle",
```
dustinburson, line 2: The test data directory and file names aren't all that clear on the intended test case.  Should we split by input / output types? From FHIR or To FHIR?

src/Microsoft.Health.Fhir.Liquid.Converter/Models/Hl7v2/MinimalHl7v2Message.cs
wi-y: We technically use the existing classes with some minor modifications so I've the existing models to have a segment name for Hl7v2Segment and a default constructor for Hl7v2Field, and then deleted the other models I had created. The reason for diverging in the first place was because the constructor of Hl7v2Field required the user to supply a list of components but for our use case it would be nice to just supply the string value for the field rather than each component. imo mapping at the component level rather than the field level is too granular (at least for a first proof of concept)

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/DateFiltersTests.cs
```diff
@@ -306,6 +316,14 @@ public void GivenAnInvalidTimeZoneHandling_WhenFormatAsDateTime_ExceptionShouldB
             Assert.Equal(FhirConverterErrorCode.InvalidTimeZoneHandling, exception.FhirConverterErrorCode);
         }
 
+        [Theory]
+        [MemberData(nameof(GetValidDataForFhirToHl7v2Conversion))]
+        public void GivenAValidData_WhenFormatAsFhirDateTime_ConvertedHl7v2DateTimeShouldBeReturned(string input, string expectedDateTime)
+        {
+            var result = Filters.FormatAsHl7v2DateTime(input);
+            Assert.Equal(expectedDateTime, result);
+        }
```
pallar-ms, line 325: minor: could you also add the invalid test cases for expected error code.

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/DateFilters.cs
```diff
@@ -95,5 +95,30 @@ public static string Now(string input, string format = "yyyy-MM-ddTHH:mm:ss.FFFZ
...
+        {
+            if (string.IsNullOrEmpty(input))
+            {
+                return input;
+            }
+
+            if (!Enum.TryParse(timeZoneHandling, true, out TimeZoneHandlingMethod outputTimeZoneHandling))
+            {
+                throw new RenderException(FhirConverterErrorCode.InvalidTimeZoneHandling, Resources.InvalidTimeZoneHandling);
+            }
```
pallar-ms, line 109: nit: This block (before line 121) seems to be common across the other 3 functions too (with the datatimetype being diff for 1 of them) - can we pull it out into a common method.

src/Microsoft.Health.Fhir.Liquid.Converter/Models/PartialDateTime.cs
```diff
@@ -129,5 +129,39 @@ public string ToFhirString(TimeZoneHandlingMethod timeZoneHandling = TimeZoneHan
...
+            }
+
+            string dateTimeFormat;
+            if (Precision < DateTimePrecision.Milliseconds)
+            {
+                dateTimeFormat = "yyyyMMddHHmmssK";
+            }
+            else
+            {
+                dateTimeFormat = MillisecondString == null ? "yyyyMMddHHmmssK" : $"yyyyMMddHHmmss{MillisecondString}K";
```
pallar-ms, line 161: nit: can combine ` || MillisecondString == null` with above check for readability, since they return the same value. Also, just in case this format was updated for one but missed for the other by mistake. But feel free to 'won't fix' if you feel this is more readable as is.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/JsonToHl7v2Processor.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        public Hl7v2Data GenerateHL7Message(JObject transformations)
+        {
+            var hl7Segments = new List<Hl7v2Segment>();
+
+            // validate that the template has a messageDefinition section
+            if (transformations["messageDefinition"] == null)
+            {
+                throw new RenderException(
+                    FhirConverterErrorCode.TemplateRenderingError,
```
pallar-ms, line 67: nit: would the PropertyNotFound code be more appropriate here?

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/JsonToHl7v2Processor.cs
```diff
@@ -0,0 +1,179 @@
...
+                    "The messageDefinition section was not found in the conversion template");
+            }
+
+            // create HL7v2 segments
+            for (int i = 0; i < transformations["messageDefinition"].Count(); i++)
+            {
+                var dataResolution = transformations["messageDefinition"][i] as JObject;
+                var segmentName = dataResolution.Properties().First().Name;
```
pallar-ms, line 75: Are we guaranteed non empty values for the properties collection?

pallar-ms: nit: save dataResolution.Properties().First() and reuse below to avoid lookup.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/JsonToHl7v2Processor.cs
```diff
@@ -0,0 +1,179 @@
...
+            var context = new JSchemaContext(
+                environments: new List<Hash> { Hash.FromDictionary(data) },
+                outerScope: new Hash(),
+                registers: Hash.FromDictionary(new Dictionary<string, object> { { "file_system", templateProvider.GetTemplateFileSystem() } }),
+                errorsOutputMode: ErrorsOutputMode.Rethrow,
+                maxIterations: Settings.MaxIterations,
+                formatProvider: CultureInfo.InvariantCulture,
+                cancellationToken: cancellationToken)
+            {
+                ValidateSchemas = new List<JsonSchema>(),
```
pallar-ms, line 162: Just out of curiosity, I've not yet figured out what this ( ValidateSchemas = new List<JsonSchema>(),) is needed for, I 've only seen this be set for the ToJsonProcessor and was wondering if you happen to know more about this.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/JsonToHl7v2Processor.cs
```diff
@@ -0,0 +1,179 @@
...
+
+                var fields = new List<Hl7v2Field>();
+
+                JObject mappingsForSegment = dataResolution.Properties().First().Value as JObject;
+
+                // Fill in any gaps in fields for the given segment
+                //   - For example if PID-3 is populated but not PID-1 or PID-2 then we should fill in placeholders up until and including PID-3
+                //   - Get highest field position in each segment and fill in lower fields with blank fields
+                int highestField = 0;
+                foreach (var mapping in mappingsForSegment)
```
pallar-ms, line 85: Just for my learning, I will probably reach out to you offline to learn more about this part, in particular, I think an example of what mappingsForSegment looks like would be super helpful. 
Also, it's pretty cool that you were able to figure out Hl7v2 format in depth and even write a liquid template from scratch!

src/Microsoft.Health.Fhir.Liquid.Converter/Models/PartialDateTime.cs
wi-y: thanks, updated

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/JsonToHl7v2Processor.cs
wi-y: that makes sense. updated

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/JsonToHl7v2Processor.cs
wi-y: no but I've added some validation code so that it will throw an exception in invalid cases

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/DateFiltersTests.cs
wi-y: added

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/JsonToHl7v2Processor.cs
wi-y: I don't know much about it. it seems like the JSON processors support some kind of schema validation on the output but I am not too familiar with it: https://github.com/microsoft/FHIR-Converter/tree/b06f905598b1cda9abc0818f6d9c7331aa5657fe/src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/TestData/TestTemplates/ValidValidateTemplates

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Processors/ProcessorTests.cs
```diff
@@ -362,5 +366,45 @@ public void GivenJObjectInput_WhenConvertWithJsonProcessor_CorrectResultShouldBe
...
+        public void GivenJObjectInput_WhenConvertWithJsonToHl7v2Processor_CorrectResultShouldBeReturned()
+        {
+            var processor = new JsonToHl7v2Processor(_processorSettings);
+            var templateProvider = new TemplateProvider(TestConstants.TestTemplateDirectory, DataType.Json);
+            var testData = JObject.Parse(_fhirBundleTestData);
+            var result = processor.Convert(testData, "BundleToHl7v2", templateProvider);
+            Assert.Equal(_fhirBundleExpectedData, result);
+        }
+
+        [Fact]
```
pallar-ms, line 380: nit suggestion: It seems like the test setup and assertion is similar for these 3 tests where exception is expected. Maybe making this a Theory test and passing the file name and expected error code will help reduce the # of lines, but if you want to have separate test case names then fine to leave as is.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/JsonToHl7v2Processor.cs
```diff
@@ -60,31 +61,89 @@ public Hl7v2Data GenerateHL7Message(JObject transformations)
...
+                {
+                    segmentName = dataResolutionForSegment.Properties().First().Name;
+                    segmentName = segmentName.ToUpper();
+                }
+                catch
+                {
+                    throw new RenderException(
+                        FhirConverterErrorCode.TemplateSyntaxError,
+                        $"The segment name in position {i} specified in the messageDefinition does not exist or is invalid");
+                }
```
pallar-ms, line 95: nit: Generally using if/else to evaluate null is more efficient than generating exceptions. It sounds like in this case, we are expecting null ref as the only exception, so I think checking for that instead (using ?. to safely access in case of null) as you have above is better.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/JsonToHl7v2Processor.cs
```diff
@@ -60,31 +61,89 @@ public Hl7v2Data GenerateHL7Message(JObject transformations)
...
 
-                JObject mappingsForSegment = dataResolution.Properties().First().Value as JObject;
+                JObject mappingsForSegment = null;
+                try
+                {
+                    mappingsForSegment = dataResolutionForSegment.Properties().First().Value as JObject;
+
+                    if (mappingsForSegment == null)
+                    {
+                        throw new Exception("The segment mapping for the template is invalid");
```
pallar-ms, line 112: why not just throw the RenderException from here itself?

pallar-ms: Same for below

pallar-ms (approved):
Left a couple suggestions.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/JsonToHl7v2Processor.cs
wi-y: there are a couple of conditions that I wanted to catch like the casting error (property exists but cannot be casted to a JObject) or the property not existing at all. in the other case there were also 2 conditions I wanted to catch like unable to parse an int, and the int being negative.

PR #501, opened 2023-11-20
Fix template references for Id/PractitionerRole (case sensitive)

shabanlushaj:
### Issue:

The current implementation of RDE feed conversion is encountering a failure due to incorrect referencing of ID/PractitionerRole template.

### Fix:
Updated template for every reference of ID/PractitionerRole, resolving the conversion failure.

### Test:

Using [Fhir Converter sample](https://github.com/microsoft/FHIR-Converter/blob/main/data/SampleData/Hl7v2/RDE-O11-01.hl7)

POST azure-server/$convert-data
<details>
  <summary>Body</summary>

```json

  {
      "resourceType": "Parameters",
      "parameter": [
          {
              "name": "rootTemplate",
              "valueString": "RDE_O11"
          },
          {
              "name": "templateCollectionReference",
              "valueString": "microsofthealth/fhirconverter:default"
          },
          
          {
              "name": "inputData",
              "valueString": "MSH|^~\\&|IMMAPP|GHHSFacility^2.16.840.1.122848.1.30^ISO|EHRApp^1.Edu^ISO|GHHRFacility^2.16.840.1.1122848.1.32^ISO|202108181126+0215|SECURITY|RDE^O11^RDE_O11|MSG00001|P|2.8|||||USA||en-US|||22 GHH Inc.|23 GHH Inc.|24GHH^2.16.840.1.114884.10.20^ISO|25GHH^2.16.840.1.114884.10.23^ISO
  SFT|Orion|2.4.3.52854|Rhapsody|2.4.3|Testactivity|20070725111624
[...]

PR #509, opened 2023-12-06
Bug fixes, default template provider, refactor functional tests

pallar-ms:
Changes include:
* Fix bug to support multiple data type template conversion using storage account blob provider.
* Add support for default template provider.
* Add UTs and functional tests.
* Refactor functional tests to test template collection providers too.


src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/ConvertDataTemplateDirectoryProviderFunctionalTests.cs
```diff
@@ -0,0 +1,128 @@
...
+using Microsoft.Health.Fhir.Liquid.Converter.Models;
+using Microsoft.Health.Fhir.Liquid.Converter.Models.Hl7v2;
+using Microsoft.Health.Fhir.Liquid.Converter.Processors;
+using Newtonsoft.Json;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests
+{
+    public class ConvertDataTemplateDirectoryProviderFunctionalTests : BaseConvertDataFunctionalTests
```
dustinburson, line 20: Not sure I understand how the new categories are being applied.  For example why is the test "GivenCcdaMessageForTimezoneTesting_WhenConvert_ExpectedResultShouldBeReturned" in this file now?

src/Microsoft.Health.Fhir.TemplateManagement/Models/DefaultTemplateInfo.cs
```diff
@@ -18,16 +18,24 @@ public DefaultTemplateInfo(DataType dataType, string imageReference, string temp
             TemplatePath = templatePath;
         }
 
+        public static string Hl7v2DefaultTemplatesResource => "Hl7v2DefaultTemplates.tar.gz";
```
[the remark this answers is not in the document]
dustinburson: Not sure I follow.  Is the ask to ensure there is a package for this code that has the default templates embedded as resources or are you suggesting the default templates be packaged in a nuget in a manner besides an embedded resource?

data/Templates/Hl7v2/Resource/_MedicationRequest.liquid
```diff
@@ -589,27 +589,27 @@
...
-                    {% evaluate practitionerRole_ID_RXO_14 using 'ID/practitionerRole' XCN: RXO.14.Repeats[0] -%}
+                    {% evaluate practitionerRole_ID_RXO_14 using 'ID/PractitionerRole' XCN: RXO.14.Repeats[0] -%}
                     "reference":"PractitionerRole/{{ practitionerRole_ID_RXO_14 }}",
                 {% elsif RXE.13.Repeats[0] %}
-                    {% evaluate practitionerRole_ID_RXE_13 using 'ID/practitionerRole' XCN: RXE.13.Repeats[0] -%}
+                    {% evaluate practitionerRole_ID_RXE_13 using 'ID/PractitionerRole' XCN: RXE.13.Repeats[0] -%}
                     "reference":"PractitionerRole/{{ practitionerRole_ID_RXE_13 }}"
                 {% elsif ORC.21.Repeats[0] %}
-                    {% evaluate practitionerRole_ID_ORC_21 using 'ID/practitionerRole' XCN: ORC.21.Repeats[0] -%}
+                    {% evaluate practitionerRole_ID_ORC_21 using 'ID/PractitionerRole' XCN: ORC.21.Repeats[0] -%}
```
dustinburson, line 598: Is there a reason to not take this PR with these changes? https://github.com/microsoft/FHIR-Converter/pull/501/files

src/Microsoft.Health.Fhir.TemplateManagement/Factory/IConvertDataTemplateCollectionProviderFactory.cs
```diff
@@ -3,12 +3,15 @@
...
 
 namespace Microsoft.Health.Fhir.TemplateManagement.Factory
 {
-    public interface ITemplateProviderFactory
+    public interface IConvertDataTemplateCollectionProviderFactory
     {
-        public IConvertDataTemplateProvider GetTemplateProvider();
+        public IConvertDataTemplateCollectionProvider CreateTemplateCollectionProvider();
+
+        public IConvertDataTemplateCollectionProvider CreateTemplateCollectionProvider(BlobContainerClient blobContainerClient);
```
wi-y, line 15: for `CreateTemplateCollectionProvider(BlobContainerClient blobContainerClient);` it seems like this will require the caller to supply the blobContainerClient. I am not opposed to this but it seems like the caller will need to decide whether to call `CreateTemplateCollectionProvider();` or `CreateTemplateCollectionProvider(BlobContainerClient blobContainerClient);` which means some template collection provider logic and Options may need to reside in the caller rather than in the factory

wi-y (approved):

src/Microsoft.Health.Fhir.TemplateManagement/Models/DefaultTemplateInfo.cs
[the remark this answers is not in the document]
pallar-ms: The templates are already packaged as embedded resources (after compression), I was wondering if we should also add these .tar.gz as content files within the nuget package additionally.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/ConvertDataTemplateDirectoryProviderFunctionalTests.cs
pallar-ms: I basically wanted to reuse a lot of the tests in the original FunctionalTests.cs file (which was previously testing against template directory) to also test using template collection provider. 
So, this file corresponds to the previous FunctionalTests.cs and the new base class has the copied common functions that can be used across both.
Specifically, the test you mentioned was always in this test class, and I didn't take it over as a common test since it seemed like the logic being tested is agnostic to whether template directory or template collection is used.

data/Templates/Hl7v2/Resource/_MedicationRequest.liquid
pallar-ms: No reason, I wasn't aware of this PR before I had already added this change, since it was required for my tests to pass.

src/Microsoft.Health.Fhir.TemplateManagement/Factory/IConvertDataTemplateCollectionProviderFactory.cs
pallar-ms: In a way, it is the same expectation as before, where the caller needed to decide whether to instantiate StorageTemplateCollectionProviderFactory() or DefaultTemplateCollectionProviderFactory() or future template collection collection providers. 
But yeah if we don't want that, we can have a generic TemplateCollectionConfiguration (maybe reuse the existing one) and depending on what options are set in there, the factory class can auto create the provider accordingly. But we would need to have some order of precedence in case the configuration had multiple provider info supplied (e.g. the config supplies both storage blob or github (future provider)). 
Thoughts?

src/Microsoft.Health.Fhir.TemplateManagement/Models/DefaultTemplateInfo.cs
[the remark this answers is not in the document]
dustinburson: Thanks for the clarification, since the Nugets are just for our use currently I am not sure there is much value in that change unless there was a specific use you were thinking of.  If we release the Nugets on Nuget.org then I think this would be a valuable change (though potentially should be in its own assembly).

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/ConvertDataTemplateDirectoryProviderFunctionalTests.cs
dustinburson: Okay, my follow up question is around having functional tests per template type, like CCDA tests specific to the functionality being tested in the test case.  It seems like we would benefit from getting more specific in the long run.

data/Templates/Hl7v2/Resource/_MedicationRequest.liquid
dustinburson: Okay, if possible it would be good to accept the original PRs so the original submitters get credit for their efforts.

src/Microsoft.Health.Fhir.TemplateManagement/Factory/IConvertDataTemplateCollectionProviderFactory.cs
dustinburson: I do like not requiring a BlobContainerClient for the template collection provider if the intent to is to use this same provider when we add other options like Github in the future.

The ideal setup in my mind would be:
* An optional configuration class to control which providers are configured.
* A factory class.  The generic one would have the configuration as a constructor param injected via DI though we could also have versions of the factory that are specific to certain cases.  For example there could be an ACR version of the factory that is just for backwards compat with the existing FHIR service.
* For the actual IConvertDataTemplateCollectionProvider implementation, once we have a setup that allows multiple sources the logic for how they are resolved and the order can be internalized inside the implementation.  It could be controlled by options in the configuration that the factory will then pass along to appropriate collection provider that is being created.

All this is a long way of saying, I do believe we should try to avoid the explicit reference to the blob client in the factory interface signature since there could be situations where a blob source isn't valid (specifically when supporting backwards compatibility).  I know the old signature with out the blob client exists but as stated above I believe we can hide this from the create call with some tweaks.

dustinburson (changes requested):
Based on the discussion thread for ```IConvertDataTemplateCollectionProvider CreateTemplateCollectionProvider(BlobContainerClient blobContainerClient);``` I think it would be good to remove this method signature if possible.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/ConvertDataTemplateDirectoryProviderFunctionalTests.cs
pallar-ms: There are tests per template type in these tests, are you looking for something specific? Those are the common ones I pulled out into the base class so we can test when using both - a TemplateCollectionProvider as well as TemplateDirectoryProvider.

src/Microsoft.Health.Fhir.TemplateManagement/Factory/IConvertDataTemplateCollectionProviderFactory.cs
pallar-ms: Yeah I did originally consider taking the container inputs as configuration options before, but since we were previously passing in the BlobContainerClient to the original StorageTemplateProviderFactory(), I left that as is as the expectation from the consumer. 
But agreed, I'll probably add a new 'StorageConfigurationOptions' class and have an optional property within TemplateCollectionConfiguration (which we already pass into this generic factory) and then internally create the appropriate providers depending on if this prop is set.

Personally, I don't like the idea of specific versions of factory, say each for ACR, Storage, Default, GitHub, etc. since I feel it could defeat the purpose of a factory class. Having a generic one that is responsible to create different template collection providers depending on provided inputs abstracts the creation logic without exposing to the client the different providers.

Also to clarify, this does not incorporate the previous ACR implementation. Prior to this change, it appeared that we have already created a segregation between the previous and new convert impl (e.g., creating new template collection provider and not extending the previous one), so going by that this new factory class doesn't deal with the prior providers.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/ConvertDataTemplateDirectoryProviderFunctionalTests.cs
dustinburson: Sorry for not being more clear, I was asking if we should have dedicated test files/directories for the different template types rather than combining them into one file.  I don't think it should impact this PR.  We can do a subsequent update if that is change we decide to make.

src/Microsoft.Health.Fhir.TemplateManagement/Factory/IConvertDataTemplateCollectionProviderFactory.cs
dustinburson: On the ACR implementation being separate, that is fair though we may want to reconsider that in the future it would be nice if the new interface could incorporate that scenario if necessary.

Multiple factories isn't needed and since right now the scenario is only for storage one implementation is fine.  If we prefer a model where factory contains all the possible sources and is driven by a config that is fine , though I do think there a case for dedicated scenario factories for each template source and there is common factory the invokes the correct ones based on the configuration.

src/Microsoft.Health.Fhir.TemplateManagement/Models/DefaultTemplateInfo.cs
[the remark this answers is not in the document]
pallar-ms: Yeah makes sense to consider when releasing the nuget publicly. Thanks

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/ConvertDataTemplateDirectoryProviderFunctionalTests.cs
pallar-ms: Gotcha, yeah agreed, I've also noticed the need for similar refactoring to improve these tests (and others in this repo) generally, that we can chip away at as and when we make changes to the relevant classes or as part of backlog. For now, given I was making changes to the template collection provider, I wanted to get test coverage on that so added in these tests (and given this is already a large PR didn't want to include more changes yet), but yes incrementally we should continue to improve the organization of these tests.

src/Microsoft.Health.Fhir.TemplateManagement/Factory/IConvertDataTemplateCollectionProviderFactory.cs
pallar-ms: As discussed, incorporated the options to figure out which provider to return and internally creating the blob container client. 
@wi-y - as part of this I moved the relevant configuration classes from the convert repo here so the factory can just read these directly instead of the consumer parsing the config and creating the client. Let me know if there are any concerns.

wi-y (dismissed):

pallar-ms:
I'm not sure why suddenly for this PR the check for 'microsoft.FHIR-Converter' is required - a pipeline that has been disable for 2-3 weeks now. It hasn't appeared in prior PRs so not sure what is causing this yet. Only the FHIR-Converter OSS PR Build is expected as a required check.

wi-y (approved):

dustinburson (approved):

PR #521, opened 2024-01-25
Refactor ConvertProcessorFactory and add UnitTests

ShaunDonn2:
Remove ConvertTelemetryConfiguration class and allow code user to pass in ProcessorSettings, so they can configure the EnableTelemetryLogging field.


ms-teli (dismissed):

ms-teli (commented):
One more thing, can we move the factor to processors folder to keep it simple?

ms-teli (approved):

PR #525, opened 2024-02-16
Generate SBOM before publish task

wi-y:


ms-teli (approved):

build/package-artifacts.yml
```diff
@@ -45,15 +45,15 @@ steps:
 
   - template: package-templates.yml
 
-# Publish artifacts
+# Generate SBOM
+  - task: ManifestGeneratorTask@0
+    displayName: 'SBOM Manifest Generator'
+    inputs:
+      BuildDropPath: '$(Build.ArtifactStagingDirectory)'
```
ms-teli, line 52: I see '/bin' is being missed here. I'm fine if it was intentional and tested.

wi-y: yeah it was intentional

PR #527, opened 2024-02-24
Use default template provider if containerUrl is empty

pallar-ms:
Currently, if containerUrl is empty, the factory attempts to create blob template provider and throws on null containerUrl. 
Instead it will be better to use default template provider and not throw on initialization.


wi-y (dismissed):
the unit test GivenInvalidStorageAccountConfiguration_WhenCreateTemplateCollectionProvider_ThenExceptionThrown may need to be updated

ShaunDonn2 (approved):

ShaunDonn2 (approved):

PR #532, opened 2024-05-01
Blob template collection size limit restriction

pallar-ms:
Check blob template collection is within the max size limit (10MB) to constrain template sizes.

### DIFF
--- a/src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateCollectionProvider.cs
+++ b/src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateCollectionProvider.cs
@@ -13,9 +13,12 @@ using Azure;
 using Azure.Storage.Blobs;
 using Azure.Storage.Blobs.Models;
 using DotLiquid;
+using EnsureThat;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Health.Fhir.Liquid.Converter.Utilities;
 using Microsoft.Health.Fhir.TemplateManagement.Configurations;
+using Microsoft.Health.Fhir.TemplateManagement.Exceptions;
+using Microsoft.Health.Fhir.TemplateManagement.Models;
 using Polly;
 using Polly.Retry;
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateCollectionProvider.cs
```diff
@@ -97,18 +104,27 @@ public BlobTemplateCollectionProvider(BlobContainerClient blobContainerClient, I
...
             {
                 foreach (BlobItem blobItem in blobPage.Values)
                 {
                     blobs.Add(blobItem.Name);
+
+                    totalBlobsSize += blobItem.Properties.ContentLength ?? 0;
+
+                    if (totalBlobsSize > _maxTemplateCollectionSizeInBytes)
+                    {
+                        throw new TemplateCollectionExceedsSizeLimitException(TemplateManagementErrorCode.BlobTemplateCollectionTooLarge, $"Total blob template collection size is larger than the size limit: {_templateCollectionConfiguration.TemplateCollectionSizeLimitMegabytes}MB.");
```
wi-y, line 126: from a user perspective I'm assuming they will see this exception if they have access to the logs. what will be the state of the container app after this exception is thrown?

wi-y: never mind - checked out the other PR. the exception is being caught and handled there

wi-y (approved):

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/BlobTemplateCollectionProvider.cs
pallar-ms: Right now since the templates are loaded when the first request comes in, the api response will indicate this error message. 
When we change that to load templates on say app startup, then it will show in the container app logs and should fail the init of the app.

PR #534, opened 2024-05-06
First draft of convert gen2 docs

pallar-ms:
First draft of convert gen2 docs


pallar-ms:
@dustinburson , @ShaunDonn2 , @wi-y  - I'm going to merge this to the shared doc branch shared/convert-api-documentation-update to unblock you all from adding content to these pages created, with the expectation this will be reviewed when we merge to main and until then we can iterate on this branch.

PR #540, opened 2024-05-08
Add monitoring/troubleshooting/bicep deploy docs

ShaunDonn2:


docs/how-to-guides/deployment-options.md
```diff
@@ -38,6 +38,66 @@ To learn more about the various options available to customize your service, and
 
 #### Option 2: Deploy a single Bicep file locally
```
pallar-ms, line 39: Can you also help fill out basics of pre-reqs, e.g., az login, az set subscription

docs/how-to-guides/deployment-options.md
```diff
@@ -38,6 +38,66 @@ To learn more about the various options available to customize your service, and
...
+Deploy the [Single Deploy Bicep Template](../deploy/FhirConverter-SingleAzureDeploy.bicep) by running the following command:
+
+```
+az deployment sub create --location <Location> --template-file FhirConverter-SingleAzureDeploy.bicep
+```
+
+Note: See [region availability](https://azure.microsoft.com/en-us/explore/global-infrastructure/products-by-region/?products=monitor,storage,container-apps) for the required resources to select a valid location for the resources to be deployed in.
+
+You will need to provide a *serviceName* that will be used to generate a name for each of the resources provisioned, and the *containerAppImageTag*, which is the tag of the FHIR Converter image version to be pulled from MCR. To see available image tags, visit the [FHIR Converter MCR page](https://mcr.microsoft.com/en-us/product/healthcareapis/fhir-converter/tags).
+
```
pallar-ms, line 50: This is fine for now, but let's consider creating a tabular format for parameters.

docs/how-to-guides/troubleshoot.md
```diff
@@ -0,0 +1,174 @@
...
+ 
+Application logs can provide detailed debugging information specific to the error. Error logs conform to the following format:
+ 
+```
+Failed conversion request with HTTP Status {HTTP status}; TOP-LEVEL-ERROR: {error code} - {top-level error message}.; INNER ERROR: Code: {inner error code}, Message: {inner error code}, Exception: {the outer exception that causes the Inner Error}; InnerException: {the inner exception of the Inner Error} - Code: {error code of the inner exception}, Message: {inner exception message}.
+```
+ 
+The top-level error information in the error log aligns with the top-level error fields in the response body. If the inner error details are present in the response body, they align with the inner error details of the log message, though the log message will likely provide more details than the response body. See the [Monitoring](monitoring.md) document for more information on accessing these logs and sample queries that may be helpful for debugging.
+ 
+## Top-Level Error Codes
```
pallar-ms, line 32: Suggestion: I think a tabular format might make this more user friendly. Especially since the columns are the same, i.e., Top level, inner error, root cause, troubleshooting

PR #545, opened 2024-05-09
Fix path returned for FHIR to HL7v2 default template errors

wi-y:
Previously when a default template did not exist the code was returning the default path of "Hl7v2/<templateName>" when converting from FHIR to HL7v2. The path should be "Fhir/<templateName>" so the PR updates the path that is returned.

### DIFF
--- a/src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirToHl7v2Processor.cs
+++ b/src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirToHl7v2Processor.cs
@@ -23,16 +23,18 @@ using NJsonSchema;
 
 namespace Microsoft.Health.Fhir.Liquid.Converter.Processors
 {
-    public class JsonToHl7v2Processor : BaseProcessor
+    public class FhirToHl7v2Processor : BaseProcessor
     {
         private readonly IDataParser _parser = new JsonDataParser();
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirToHl7v2Processor.cs
```diff
@@ -23,17 +23,19 @@
...
 
         private string[] _segmentsWithFieldSeparator = new string[] { "MSH", "BHS", "FHS" };
 
-        public JsonToHl7v2Processor(ProcessorSettings processorSettings, ILogger<JsonToHl7v2Processor> logger)
+        public FhirToHl7v2Processor(ProcessorSettings processorSettings, ILogger<FhirToHl7v2Processor> logger)
             : base(processorSettings, logger)
         {
         }
 
+        protected override DataType DataType { get; set; } = DataType.Fhir;
```
pallar-ms, line 37: I don't think this would work as is. The value in DataType is used as the base folder within which we extract and check the default templates, and currently, DataType.Fhir is also used for FhirProcessor (STU3), so this will conflict.

pallar-ms: I think when you add the default template for Fhir to Hl7v2, you will run into this. It might end up extracting both STU3 and the new templates you'll be adding into the same folder named 'Fhir', and cause issues on template lookups.

PR #547, opened 2024-05-10
Addressing bug bash feedback on deployment scripts

ShaunDonn2:
- update role assignment from storage account to storage account container
- add timestamp to deployment names
- change app insights connection string env var to be kv secret reference
- update ps1 script with bicep changes
- add ARM template


docs/deploy/Deploy-DependentResources.bicep
```diff
@@ -0,0 +1,124 @@
...
+  'ukwest'
+  'westcentralus'
+  'westeurope'
+  'westus'
+  'westus2'
+  'westus3'
+])
+param location string
+
+@description('If set to true, a storage account and container will be deployed with the specified names for storing custom templates.')
```
pallar-ms, line 54: super nit: `blob container` instead.

docs/deploy/Deploy-DependentResources.bicep
```diff
@@ -0,0 +1,124 @@
...
+
+resource keyVault 'Microsoft.KeyVault/vaults@2021-04-01-preview' = if (deployKeyVault) {
+  name: keyVaultName
+  location: location
+  properties: {
+    sku: {
+      family: 'A'
+      name: 'standard'
+    }
+    tenantId: subscription().tenantId
```
pallar-ms, line 100: nit: fix indentation

docs/deploy/Deploy-FhirConverterService.bicep
```diff
@@ -72,22 +84,28 @@ param imageTag string
...
+param applicationInsightsConnectionStringSecretName string = ''
 
-@description('The resource ID of the user-assigned managed identity used to access the application insights instance.')
-param applicationInsightsUAMIResourceId string = ''
+// Get the container apps environment
+resource containerAppsEnvironment 'Microsoft.App/managedEnvironments@2023-05-01' existing = {
+  name: envName
+}
 
-@description('The ID of the container apps environment where the container app should be deployed to.')
```
pallar-ms, line 84: Why is this reverted?

docs/deploy/Deploy-FhirConverterService.bicep
```diff
@@ -108,53 +126,72 @@ var securityAuthenticationAudiencesConfig = [for (audience, i) in securityAuthen
 }]
 
 var securityConfiguration = concat(securityEnabledConfiguration, securityEnabled ? concat(securityAuthenticationAuthorityConfig, securityAuthenticationAudiencesConfig) : [])
+var integrateTemplateStore = !empty(templateStorageAccountName) && !empty(templateStorageAccountContainerName)
```
pallar-ms, line 129: nit: move this to co-locate under Template hosting configuration.

docs/deploy/Deploy-FhirConverterService.bicep
```diff
@@ -108,53 +126,72 @@ var securityAuthenticationAudiencesConfig = [for (audience, i) in securityAuthen
...
+var envConfiguration =  concat(securityConfiguration, telemetryConfiguration, blobTemplateHostingConfiguration)
 
 var imageName = 'healthcareapis/fhir-converter'
 
+// Configure identities
+var applicationInsightsUAMIResourceId = configureApplicationInsights ? applicationInsightsUAMI.id : ''
+var keyVaultUAMIResourceId = resourceId('Microsoft.ManagedIdentity/userAssignedIdentities', keyVaultUAMIName)
+var userAssignedIdentities = configureApplicationInsights ? {
+  '${applicationInsightsUAMIResourceId}' : {}
+  '${keyVaultUAMIResourceId}' : {}
```
pallar-ms, line 166: Adding `keyVaultUAMIResourceId` should be decoupled from configureAppInsights and based off if the value was provided. Fine for now to leave, but in the future if we add another configuration which requires a secret, this will need to change.

docs/deploy/Deploy-FhirConverterService.ps1
```diff
@@ -70,13 +66,23 @@ param
...
+    [string]$keyVaultName = "$($serviceName)-kv",
 
-    [string[]]$securityAuthenticationAudiences = @(),
+    [string]$keyVaultUserAssignedIdentityName = "$($serviceName)-kv-identity",
 
-    [string]$securityAuthenticationAuthority = "",
+    [bool]$securityEnabled = $false,
 
-    [bool]$deployApplicationInsights = $true
+    [string[]]$securityAuthenticationAudiences,
```
pallar-ms, line 83: Curious - why remove the default values?

docs/deploy/Deploy-Infrastructure.bicep
```diff
@@ -54,14 +65,48 @@ resource logAnalyticsWorkspace 'Microsoft.OperationalInsights/workspaces@2020-03
...
+resource keyVault 'Microsoft.KeyVault/vaults@2021-04-01-preview' existing = if (deployApplicationInsights) {
+  name: keyVaultName
+}
+
+var applicationInsightsConnectionStringSecretName = '${applicationInsightsName}-conn-string'
+resource applicationInsightsConnectionStringSecret 'Microsoft.KeyVault/vaults/secrets@2021-04-01-preview' = if (deployApplicationInsights) {
+  parent: keyVault
+  name: applicationInsightsConnectionStringSecretName
+  properties: {
+    value: deployApplicationInsights ? applicationInsights.properties.ConnectionString : 'default'
```
pallar-ms, line 90: Isn't this check redundant here if we are already checking when deploying this resource?

ShaunDonn2: I thought so, but it seems that the value is still evaluated even when we don't deploy the resource, so it was causing a resource not found error in cases where the bool was false. Adding this check resolved the issue.

docs/deploy/Deploy-DependentResources.bicep
```diff
@@ -0,0 +1,124 @@
...
+  'westus2'
+  'westus3'
+])
+param location string
+
+@description('If set to true, a storage account and blob container will be deployed with the specified names for storing custom templates.')
+param deployTemplateStore bool
+
+@description('Name of the storage account to be deployed.')
+param templateStorageAccountName string
```
pallar-ms, line 58: If `deployTemplateStore` is set to false, doesn't make sense to require the names so we should have empty defaults.

pallar-ms (approved):

PR #549, opened 2024-05-10
Deployment doc update

ShaunDonn2:


docs/how-to-guides/deployment-options.md
```diff
@@ -51,78 +52,58 @@ Note: You are also able to update the service configuration post initial deploym
 
 #### Default settings
 
-(**TODO** mention default deployment settings, i.e., auth disabled but recommend enabling, template store created, app insights created, etc.)
+* **Security** - Security settings for the API endpoints are disabled by default. It is **strongly recommended** to enable security for your FHIR Converter service.
+* **Template store integration** - Template store integration is disabled by default. When template store integration is disabled, a storage account will not be provisioned with the deployment and the FHIR converter service will use the provided default templates. To use custom templates, template store integration must be enabled so that custom templates can be stored in the deployed storage account.
```
pallar-ms, line 56: maybe also a line that the service will use default templates in this scenario.

docs/how-to-guides/deployment-options.md
```diff
@@ -51,78 +52,58 @@ Note: You are also able to update the service configuration post initial deploym
...
+| templateStoreIntegrationEnabled | boolean | No | Enable or disable template store integration for the FHIR Converter service. If disabled, only default templates can be used for conversion. If enabled, only custom templates can be used. | false |
+| templateStorageAccountName | string | No | The name of the storage account to store custom templates. | serviceName + "templatestorage" |
+| templateStorageAccountContainerName | string | No | The name of the blob container to store custom templates. | serviceName + "templatecontainer" |
+| applicationInsightsEnabled | boolean | No | Enable or disable Application Insights deployment for your service. Application metrics and request logs will be unavailable for any time period where Application Insights is disabled. | true |
+| keyVaultName | string | No | The name of the Key Vault to store secrets. | serviceName + "-kv" |
+| keyVaultUserAssignedIdentityName | string | No | The name of the user-assigned managed identity to be used to access secrets in the Key Vault. | serviceName + "-kv-identity" |
+| securityEnabled | boolean | No | Enable or disable security settings for the APIs. | false |
```
pallar-ms, line 84: This is great! 

One suggestion but please feel free to won't fix cause it is purely person to person - maybe we can put this under a separate sub section titled Parameters. Then the default settings and configurable settings might tie in with that more clearly. And the configurable settings can just be referring to the parameters if they want to change and the configuration-settings for more info.

pallar-ms: On the flip side, I can also see how having this detailed table before default settings might be too much for someone who doesn't really want the specifics at first. 
So I'll leave it up to you. :)

ShaunDonn2: Thanks, I think I prefer that flow as well. Will make the update.

pallar-ms (approved):

docs/how-to-guides/deployment-options.md
ShaunDonn2: Would it make sense to add the parameters section below "Configurable settings" so that we get the best of both?

pallar-ms: Yeah that works too!

PR #554, opened 2024-05-13
Doc updates

ShaunDonn2:
- sort error codes into http status codes
- add instructions for deploying Container App to existing env


wi-y (approved):

wi-y (dismissed):

docs/how-to-guides/deployment-options.md
```diff
@@ -133,6 +133,30 @@ This [PowerShell deployment script](../deploy/Deploy-FhirConverterService.ps1) s
...
+### Deploy a Container App to an existing Container Apps Environment
+
+To deploy a Container App to an existing Container Apps environment, use the [Deploy-FhirConvertService.bicep](../deploy/Deploy-FhirConverterService.bicep) template using the following command:
+
+```
+az deployment group create --resource-group <Resource Group containing existing Container Apps Environment> --template-file Deploy-FhirConverterService.bicep --parameters location=<Location of Container App Environment> appName=<Container App name> envName=<Name of existing Container Apps Environment> imageTag=<Image tag from MCR> 
+```
+
+In addition to the parameters listed for the [FhirConverter-SingleAzureDeploy](../deploy/FhirConverter-SingleAzureDeploy.bicep) in [Parameters](#parameters), the Deploy-FhirConvertService bicep template takes the following parameters:
+
```
pallar-ms, line 145: I'm a little confused by this section. Why are these parameters separated from the above?

docs/how-to-guides/troubleshoot.md
```diff
@@ -45,228 +45,73 @@ Application logs can provide detailed debugging information on the errors encoun
...
+| TemplateCollectionError | TemplateCollectionSizeExceedsLimit | The template collection uploaded to the Storage Account exceeds the allowed size limit. | The response body and error log should indicate the maximum allowed template collection size. Remove templates from the Storage Account so that the collection size aligns with the limit, restart the container, and then retry the request. |
```
pallar-ms, line 80: We should specify the limit here - 10MB

docs/how-to-guides/deployment-options.md
ShaunDonn2: I separated them because they are specific to this bicep template, rather than the single deploy template. I'll find a way to combine them so that it's clearer

pallar-ms: Oh got it. Can we provide a way to use the same single deploy template to setup a new service? I think we pretty much expose all the expected configs there?

ShaunDonn2: Sure, we will just need to expose these 4 additional fields. I'll add that in a separate PR so that we can get the troubleshooting changes out sooner.

pallar-ms: Ah got it, I thought we already would have those, In that case, its fine, can leave it the way you have it unless we get any feedback. Did you already make that change? Sorry for the confusion!

ShaunDonn2: Work item created: https://microsofthealth.visualstudio.com/Health/_workitems/edit/120425

wi-y (approved):

wi-y (approved):

pallar-ms (dismissed):

pallar-ms (approved):

pallar-ms (approved):

PR #557, opened 2024-05-16
Add default template for FHIR to HL7v2 conversions

wi-y:
Adds a default template that transforms a FHIR Bundle with an Observation Resource to an ORU^R01 HL7v2 message.


src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/BaseConvertDataFunctionalTests.cs
```diff
@@ -229,6 +228,20 @@ public static IEnumerable<object[]> GetDataForJson()
             });
         }
 
+        public static IEnumerable<object[]> GetDataForFhirToHl7v2()
+        {
+            var data = new List<string[]>
+            {
+                new[] { @"ObservationBundle", @"ObservationBundle.json", @"ObservationBundle-expected.hl7" },
```
pallar-ms, line 235: Should there be a case for ExamplePatient.json that you added in sample data as well?

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/BaseProcessor.cs
```diff
@@ -35,7 +34,7 @@ protected BaseProcessor(ProcessorSettings processorSettings, ILogger<BaseProcess
 
         protected virtual string DataKey { get; set; } = "msg";
 
-        protected virtual DataType DataType { get; set; }
```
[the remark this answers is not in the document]
pallar-ms: Asking cause I had used this for default root template path, but it was already added prior to that so wondering if it was used in a different context as well.

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/DefaultTemplateCollectionProvider.cs
```diff
@@ -55,8 +55,9 @@ public DefaultTemplateCollectionProvider(IMemoryCache templateCache, TemplateCol
             var ccdaDefaultTemplatesTask = Task.Run(() => templates.Add(ExtractTemplatesFromResource(DefaultTemplateInfo.CcdaDefaultTemplatesResource, DataType.Ccda.ToString())), cancellationToken);
             var jsonDefaultTemplatesTask = Task.Run(() => templates.Add(ExtractTemplatesFromResource(DefaultTemplateInfo.JsonDefaultTemplatesResource, DataType.Json.ToString())), cancellationToken);
             var stu3ToR4DefaultTemplatesTask = Task.Run(() => templates.Add(ExtractTemplatesFromResource(DefaultTemplateInfo.Stu3ToR4DefaultTemplatesResource, DataType.Fhir.ToString())), cancellationToken);
+            var fhirToHl7v2DefaultTemplatesTask = Task.Run(() => templates.Add(ExtractTemplatesFromResource(DefaultTemplateInfo.FhirToHl7v2DefaultTemplatesResource, "FhirToHl7v2")), cancellationToken);
```
pallar-ms, line 58: Is the new enum not accessible here?

src/Microsoft.Health.Fhir.Liquid.Converter/Models/PartialDateTime.cs
```diff
@@ -158,7 +158,7 @@ public string ToHl7v2Date(TimeZoneHandlingMethod timeZoneHandling = TimeZoneHand
             }
             else
             {
-                dateTimeFormat = $"yyyyMMddHHmmss{MillisecondString}K";
+                dateTimeFormat = $"yyyyMMddHHmmss.{MillisecondString}K";
```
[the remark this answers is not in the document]
wi-y: it is a breaking change, but the original function was incorrect and not conformant with HL7v2 format for milliseconds so this PR corrects the format. I had created this filter for future HL7v2 mapping templates and the sample template in this PR is the first to use it so it is probably fine to correct it.

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/BaseConvertDataFunctionalTests.cs
wi-y: It could be another example. In this PR the intent was to include the necessary FHIR resources to generate an ORU^R01 result HL7v2 message and the minimum needed for that would be a FHIR bundle of Patient and Observation, and Encounter (depending on HL7 version).

src/Microsoft.Health.Fhir.TemplateManagement/ArtifactProviders/DefaultTemplateCollectionProvider.cs
wi-y: it is - thanks - updated to use the enum

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/BaseConvertDataFunctionalTests.cs
```diff
@@ -350,5 +363,21 @@ protected void ConvertJsonMessageAndValidateExpectedResponse(ITemplateProvider t
...
+        {
+            var jsonProcessor = new FhirToHl7v2Processor(_processorSettings, FhirConverterLogging.CreateLogger<FhirToHl7v2Processor>());
+            var inputContent = File.ReadAllText(inputFile);
+            var expectedContent = File.ReadAllText(expectedFile);
+            var actualContent = jsonProcessor.Convert(inputContent, rootTemplate, templateProvider);
+
+            var fieldValues = actualContent.Split("|");
+
+            // remove MSH-7 timestamp
+            fieldValues[6] = string.Empty;
```
pallar-ms, line 377: What is the timestamp here in this context? 
Asking cause with Rob's change to handle date time parsing, we should probably validate any timestamp fields are being preserved as is without UTC conversion.

src/Microsoft.Health.Fhir.Liquid.Converter/Models/PartialDateTime.cs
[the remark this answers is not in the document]
pallar-ms: Ah gotcha, I had missed that this was inside .ToHl7v2Data() so basically doesn't impact $convert anyway.

src/Microsoft.Health.Fhir.TemplateManagement/Models/DefaultTemplateInfo.cs
```diff
@@ -36,6 +38,7 @@ public DefaultTemplateInfo(DataType dataType, string imageReference, string temp
             { "microsofthealth/ccdatemplates:default", new DefaultTemplateInfo(DataType.Ccda, "microsofthealth/ccdatemplates:default", CcdaDefaultTemplatesResource) },
             { "microsofthealth/jsontemplates:default", new DefaultTemplateInfo(DataType.Json, "microsofthealth/jsontemplates:default", JsonDefaultTemplatesResource) },
             { "microsofthealth/stu3tor4templates:default", new DefaultTemplateInfo(DataType.Fhir, "microsofthealth/stu3tor4templates:default", Stu3ToR4DefaultTemplatesResource) },
+            { "microsofthealth/fhirtohl7v2templates:default", new DefaultTemplateInfo(DataType.Fhir, "microsofthealth/fhirtohl7v2templates:default", FhirToHl7v2DefaultTemplatesResource) },
```
pallar-ms, line 41: I think adding this here will imply theoretically $convert requests can use these templates. The processing will fail since the FhirToHl7v2 processor is not present there, but does this need to be added here for the new convert api flow as well?

src/Microsoft.Health.Fhir.Liquid.Converter.FunctionalTests/BaseConvertDataFunctionalTests.cs
wi-y: MSH-7 the time that the HL7v2 message was generated. I just used the current time for populating MSH-7 which will always be changing whenever the tests run so I excluded the timestamp from validation. The timestamp isn't really used for anything important. Some systems use it for searching for messages in certain time range but it does not have any clinical impact

pallar-ms: Got it, just wanted to ensure it wasn't supposed to be a timestamp value from the input FHIR bundle/resource.

pallar-ms (dismissed):

src/Microsoft.Health.Fhir.TemplateManagement/Models/DefaultTemplateInfo.cs
wi-y: yeah the code DefaultTemplateCollectionProvider.cs is shared between the MCR fhirconverter and fhir-server and this is the only way to make the default templates work (or not work). It is pretty much all or nothing unless we wanted to modify the default template logic by runtime environment, or something like that, or have a config for the runtime that is passed in and chooses which templates to load. it is probably easiest to load all the templates and just not document the FhirToHl7v2 templates in the $convert docs

pallar-ms: But this map DefaultTemplateMap is only used in the ACR default template collection right? DefaultTemplateCollectionProvider.cs doesn't seem to be using this map specifically. Or am I missing something?

wi-y: hmm yes you are right. we can remove it from the map so it should not be able to be called from $convert, so I went ahead and removed it. The new template will still be loaded on startup but should not be able to be referenced now that it is removed from the map

pallar-ms (approved):

PR #558, opened 2024-05-16
Update build trigger to align with build exclusion list

ShaunDonn2:


build/.vsts-pr.yml
```diff
@@ -3,7 +3,14 @@
...
 
-trigger: none
+trigger:
+  paths:
+    include:
+    - '*'
+    exclude:
+    - '*.md'
+    - docs
+    - data
```
pallar-ms, line 13: I think maybe we should remove 'data' from here and ci build as there are the template management UTs that should be run if things change in data/Templates

PR #566, opened 2024-06-19
Add case for JsonException in template rendering stage

ShaunDonn2:
The Newtonsoft.Json.JsonException case is currently unhandled in the RenderTemplates method, and so it is returned as a 500 by APIs that consume this code. In this context, this exception should be the result of a poorly formed template or input data (or input data that is incompatible with the template) and so should be returned as a 400 in convert service code bases that consume this code.


src/Microsoft.Health.Fhir.Liquid.Converter/Processors/BaseProcessor.cs
```diff
@@ -155,6 +155,10 @@ protected string RenderTemplates(Template template, Context context)
             {
                 throw new RenderException(FhirConverterErrorCode.TimeoutError, Resources.TimeoutError, ex);
             }
+            catch (JsonException ex)
+            {
+                throw new RenderException(FhirConverterErrorCode.TemplateParsingError, string.Format(Resources.TemplateParsingError, ex.Message), ex);
```
pallar-ms, line 160: I'm not sure if this qualifies as a template parsing error? I would think template parsing error would occur say if the template provided is semantically incorrect as per Liquid syntax and we cannot parse it. But from what I understood of the error, the template is valid, the content is valid Hl7v2, but together they are incompatible resulting in TemplateRenderingError. 
Can you clarify?

Do we know other scenarios where TemplateRenderingError could be due to a service issue and not wrt template or incompatible data?

ShaunDonn2: That is my understanding as well. I just created the new error code "TemplateParsingError" to capture this case because it's an issue with the template parsing the input, even though they are both individually valid. I agree though, the naming could be confusing. I could name is something more specific like "IncompatibleInputAndTemplateFailedRendering".

For the TemplateRenderingError code, the only potential service issues that comes to mind, but seem very unlikely, are:
- NullArgumentExceptions can be thrown by certain methods up the call stack, but I think this is unexpected behaviour
- some sort of out of memory error, but it seems that the timeout error is hit in cases where the template or input is very large, so this might already be handled. 

The way the code is currently written, the TemplateRenderingError captures any unhandled exceptions; I am hesitant to classify all of these as 400s, in case there is some unhandled service issue that comes up that we don't anticipate. Wondering what your thoughts are on doing this, though, compared to just handling the errors we can anticipate as 400s and leaving any unhandled errors to be thrown as 500s.

pallar-ms: Yes I don't think we should make all of them 400 especially if there are service issues that could throw with this error code too. 
Having said that, will this change here not break existing users of $convert APIs who are now expecting this particular error scenario to return this TemplateRenderingError code. If we change the error code, any client consuming this package will have to handle it differently. 
I was thinking this is something we handle in the new Convert APIs based on the exception or inner exception details, and perhaps can leverage the IncompatibleInputData error code that was defined there. Thoughts?

ShaunDonn2: Yah I agree, it would break consumers of the API that expect the TemplateRenderingError for these conditions. I figured we would need to update both the new Convert APIs and the $convert-data code in the FHIR service code to handle this case to make both more correct, but then we'd have to classify that as a breaking change. If we feel that it is not worth doing that for the FHIR service $convert-data code, then definitely using the FhirConverterErrorCode of the inner exception in the new Convert APIs will work and would be a simpler solution overall, and would avoid having to impact all consumers of the OSS code. The more I think about it and given our current priorities, I am leaning more towards the solution you suggested. I'll go ahead and implement this in the new APIs instead.

PR #568, opened 2024-07-09
Enable conditional Network Isolation for template Storage Account

ShaunDonn2:
To align with the latest security recommendations, provide the option for users of the deployment scripts to protect their template Storage Account by disabling public network access and linking it to a Virtual Network. The changes also result in the Container Apps environment to be created within the same Virtual Network, so that the FHIR Converter service still has access to the Storage Account for custom template requests.


docs/deploy/FhirConverter-SingleAzureDeploy.bicep
```diff
@@ -93,12 +93,27 @@ param keyVaultUserAssignedIdentityName string = '${serviceName}-kv-identity'
...
-@description('Audiences for the api authentication.')
+@description('Audiences for the api authentication. Only applicable when securityEnabled is set to true.')
 param securityAuthenticationAudiences array = []
 
-@description('Authority for the api authentication.')
+@description('Authority for the api authentication. Only applicable when securityEnabled is set to true.')
 param securityAuthenticationAuthority string = ''
 
+@description('When set to true, the template Storage Account will only accept network traffic from within the specified Virtual Network, which the Container Apps environment will be onboarded to.')
+param storageAccountNetworkIsolationEnabled bool = false
```
pallar-ms, line 103: nit: should this be a generic param like the one you have defined in other templates 'configureNetworkIsolation' ? Since the VNet if created is tied to multiple resources ACA env, storage account, etc.

docs/how-to-guides/deployment-options.md
```diff
@@ -89,6 +89,11 @@ The table below outlines the parameters that can be configured through any of th
 | securityEnabled | boolean | No | Enable or disable security settings for the APIs. | false |
 | securityAuthenticationAudiences | array of strings | No | The audiences for the API authentication. If securityEnabled is set to true, this parameter must have a value provided. | empty array |
 | securityAuthenticationAuthority | string | No | The authority for the API authentication. If securityEnabled is set to true, this parameter must have a value provided. | empty string |
+| storageAccountNetworkIsolationEnabled | boolean | No | This is an optional security enhancement. If set to true, the template Storage Account will only be accessible to resources within a specified Virtual Network (one will be created if it doesn't already exist), and the Container Apps environment will be deployed on the Virtual Network so that it can access the Storage Account. | false |
```
pallar-ms, line 92: nit: Similar to my other comment, VNet's can be configured with ACA Env even without the storage account necessitating it, so this param could be more generic like networkIsolationEnabled perhaps.

pallar-ms: But fine to leave as is since I see below the documentation is more specifically addressing the storage account n/w isolation.

pallar-ms (approved):

docs/deploy/FhirConverter-SingleAzureDeploy.bicep
ShaunDonn2: What that said, this has me thinking about the option of isolating everything in the infrastructure (i.e. key vault, etc.) on the storage account. It's not part of our security requirement, so maybe not high priority, but do you think that would be worthwhile?

pallar-ms: No, I don't think we should address that atm if it is not being driven by security wave requirements currently.

PR #574, opened 2024-10-29
[Draft]: Add Fhir Normalization Data Type & Processor

kenziedolish:


src/Microsoft.Health.Fhir.Liquid.Converter/Models/DataType.cs
```diff
@@ -14,5 +14,6 @@ public enum DataType
         Ccda,
         Json,
         Fhir,
+        FhirNormalization,
```
pallar-ms, line 17: DataType represents the format of the data being processed. So I was thinking this would be FhirR4 or something since Fhir is already taken and is being used for STU3 conversion. IMO FhirNormalization is not data type but rather normalization is the scenario where we intend to use Fhir Converter.

PR #578, opened 2024-11-08
Migrating from dotnet 6 to dotnet 8

maryamasaaditeh:


Directory.Build.props
```diff
@@ -11,6 +11,7 @@
     <Product>Microsoft Health</Product>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <RepoRootPath>$(MSBuildThisFileDirectory)</RepoRootPath>
+	<LatestFramework>net8.0</LatestFramework>
```
kenziedolish, line 14: nit: spacing

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/DateFiltersTests.cs
```diff
@@ -50,7 +50,7 @@ public static IEnumerable<object[]> GetValidDataForAddSeconds()
             yield return new object[] { string.Empty, 60, "local", string.Empty };
             yield return new object[] { @"1970-01-01T00:01:00.000+10:00", -60, "utc", @"1969-12-31T14:00:00.000Z" };
             yield return new object[] { @"1970-01-01T00:01:00.000+05:30", -60, "utc", @"1969-12-31T18:30:00.000Z" };
-            yield return new object[] { @"1970-01-01T00:01:00Z", 60.123, "preserve", @"1970-01-01T00:02:00.123Z" };
+            yield return new object[] { @"1970-01-01T00:01:00Z", 60.123, "preserve", @"1970-01-01T00:02:00.122Z" };
```
kenziedolish, line 53: Why was this change made?

dustinburson:
Need to address build issues.  @pallar-ms do you have any suggestions on how this branch can be tested with our container prior to commiting the PR?

pallar-ms:
> Need to address build issues. @pallar-ms do you have any suggestions on how this branch can be tested with our container prior to commiting the PR?

@dustinburson , @maryamasaaditeh 

Regarding the build issue specifically, it seems to be due to the incorrect dotnet sdk version setup in the pipeline that should be fixed - https://github.com/microsoft/FHIR-Converter/blob/931a935a5bb3b9ccde4b1e9cf7396f2d94b95643/build/build.yml#L6

Generally for testing personal OSS changes in convert g2 api container, you would need to do the following:
* Create a local nuget package from your FHIR-Converter OSS branch - e.g., 
dotnet pack -p:NuspecProperties="version=1011.01" --configuration Debug/Release
* Update package reference to your local nuget source and package in your personal convert repo branch- https://microsofthealth.visualstudio.com/Health/_git/convert?path=/Directory.Packages.props&version=GBmain&line=20&lineEnd=20&lineStartColumn=4&lineEndColumn=88&lineStyle=plain&_a=contents
* Build and deploy personal convert service and infra env - https://microsofthealth.visualstudio.com/Health/_git/convert?path=/convert/deploy
* Create test convert app in your personal env - https://github.com/microsoft/FHIR-Converter/blob/main/docs/how-to-guides/deployment-options.md#deploy-a-container-app-to-an-existing-container-apps-environment

maryamasaaditeh:
> Need to address build issues. @pallar-ms do you have any suggestions on how this branch can be tested with our container prior to commiting the PR?

@dustinburson , which build issue you're referring, is there any pipeline I have to check ? since this branch is not merged yet to the main branch

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/DateFiltersTests.cs
maryamasaaditeh: It has changed to resolve one of the unit test failing

maryamasaaditeh:
All the steps mentioned for verifying the changes has been done successfully, the app container deployed : 
https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/4092fd42-145e-4e91-aa15-96f66d2b6deb/resourceGroups/hlsinteropconvertinfra-mary-wus3-01/providers/Microsoft.App/containerApps/maryconverterapp/containerapp

docs/how-to-guides/Deploy-FhirCOnverter-local.md
```diff
@@ -0,0 +1,48 @@
+# Fhir app container deployment for local changes
```
dustinburson, line 1: @maryamasaaditeh The API for the convert isn't currently in OSS so at this time there isn't any way for a external user to create a custom image at this time.  Thank you for adding these instructions but can you please remove them from the OSS repo and add them to the internal Convert repo under the docs folder?  Thank you!

dustinburson (changes requested):

docs/how-to-guides/Deploy-FhirCOnverter-local.md
maryamasaaditeh: done

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/DateFiltersTests.cs
dustinburson: Sorry, I missed this earlier. This doesn't look correct.  It seems like with the upgrade to .NET verisons the behavior has changed.  Can you please dig deeper to see if there is a way to preserve the original?

dustinburson (changes requested):

src/Microsoft.Health.Fhir.Liquid.Converter.UnitTests/Filters/DateFiltersTests.cs
maryamasaaditeh: I did some investigations and looks like the value of date time which is calculated for adding 60.123 seconds is {00:02:00.1229999} time of the day, number of ticks is also calculated : 621355969201229999 , I think that's why in test it is expecting 1970-01-01T00:02:00.122Z.
if we want to have  a unit test which both the input and result having same seconds we can use 60.124 , since for 60.124 , the value of the daytime would be something like {00:02:00.1240000}

dustinburson: So I dug a little deeper, this test is calling into:
https://github.com/microsoft/FHIR-Converter/blob/931a935a5bb3b9ccde4b1e9cf7396f2d94b95643/src/Microsoft.Health.Fhir.Liquid.Converter/Filters/DateFilters.cs#L17

which in turn is invoking:
https://github.com/microsoft/FHIR-Converter/blob/931a935a5bb3b9ccde4b1e9cf7396f2d94b95643/src/Microsoft.Health.Fhir.Liquid.Converter/Models/PartialDateTime.cs#L85

60.123 is supposed to represent 60 seconds and 123 miliiseconds so the expected output should be a minute greater and 123 milliseconds.  

I took a look at breaking changes in the .NET versions, I didn't see anything relevant in [.NET 8 ](https://learn.microsoft.com/en-us/dotnet/core/compatibility/8.0)
but I did see this in the .NET 7 breaking changes:
https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/7.0/datetime-add-precision

Digging deeper, if you use a value like 1.123 you get the expected number of milliseconds.  The value seems to deviate once you get to a second value of 8+.

To preserve the original logic I think we can modify the code in https://github.com/microsoft/FHIR-Converter/blob/931a935a5bb3b9ccde4b1e9cf7396f2d94b95643/src/Microsoft.Health.Fhir.Liquid.Converter/Models/PartialDateTime.cs#L85

from
```
DateTimeValue = DateTimeValue.AddSeconds(seconds);
```
to 
```
var timespan = TimeSpan.FromSeconds(seconds)
DateTimeValue = DateTimeValue.Add(timespan);
```

This seems to preseve the original intent.

dustinburson (approved):

Issue #468, opened 2023-04-27
CodeMapping Context should be added on the Json processor. 


guilhermekf:
Current we can use that on CCDA, and Hl7v2.
There is some reason for not allow this to be used on JSON? 
The code mapping context is a good approach to map the codes from the source to the correct Fhir value and it should be enable on the JSON processor

dbouckaert:
Is this being worked on as an enhancement?  This will make the coding a lot easier to maintain if it is in the code mapping context.

PR #591, opened 2025-03-25
Issues/468 filter getproperty for all processors

peterdanielssonservicewell:
This PR enhances the Filter GetProperty method across all processors to ensure that both code mappings **CodeSystem/CodeSystem** and **ValueSet/ValueSet** are read by default.

Resolves microsoft/FHIR-Converter#468

**Changes**

* Context Initialization
Adjusted context setup so that both default mappings are automatically loaded into the environment, ensuring they are available during property access.
* Adds support for combining multiple CodeMapping instances by introducing the CodeMapping.Append() method.
* Splits CreateContext(...) into CreateBaseContext(...) and CreateContext(...) for improved separation of context initialization logic.

**Impact**

This change improves consistency and reliability when processing filters across different processors by guaranteeing that both default mappings are always considered. There are no breaking changes expected for consumers relying on the current behavior, as the default mappings remain unchanged.


peterdanielssonservicewell:
@microsoft-github-policy-service agree company="Service Well"

PR #593, opened 2025-05-23
Initial commit of cg analysis code

rogordon01:
Adding an 'analyze' stage to perform CG and other security tasks

### DIFF
--- a/.config/tsaoptions.json
+++ b/.config/tsaoptions.json
@@ -1,0 +1,15 @@
+{
+    "instanceUrl": "https://microsofthealth.visualstudio.com",
+    "TSADocumentation": "https://aka.ms/OBTSA",
+    "projectName": "Health",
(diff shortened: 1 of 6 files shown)


.config/tsaoptions.json
```diff
@@ -0,0 +1,15 @@
+{
+    "instanceUrl": "https://microsofthealth.visualstudio.com",
+    "TSADocumentation": "https://aka.ms/OBTSA",
+    "projectName": "Health",
+    "areaPath": "Health\\HLS Interop\\Aegean",
+    "notificationAliases": [
+        "duburson@microsoft.com",
```
rogordon01, line 7: @dustinburson is there an different alias that we should use here? We have `Aegean Support <medtechsupport@microsoft.com>` and `HLS_DP_InterOp_CS@microsoft.com`

dustinburson (approved):

PR #614, opened 2026-04-10
feat: Add FHIR R4-to-R4 conversion with template variable support (Phase 1)

Chethan77:
## Summary

Implements Phase 1 of the Convert Engine Variable Support specification by adding a new FHIR R4-to-R4 conversion pathway with named string variables available at template render time.

The primary use case for this phase is Dragon Copilot MRN identifier injection: selecting a Patient identifier by `type.coding.code` and writing a new Dragon Copilot identifier with a configurable system URI.

## Scope Delivered

| Requirement | Status | Implementation |
|---|---|---|
| P1-R1: Support FHIR R4 → FHIR R4 transformation | ✅ | Added `FhirR4Processor` using the existing JSON parsing and post-processing pipeline |
| P1-R2: Add named string variables to the conversion call | ✅ | Added variable-aware overloads on `IFhirConverterWithVariables` |
| P1-R3: Variables accessible in templates | ✅ | Injected variables into Liquid context under `vars` |
| P1-R4: Create Dragon MRN template | ✅ | Added `DragonPatientMrn.liquid` |
| P1-R5: Single match enforcement | ✅ | Template raises descriptive errors for zero or multiple matches |
| P1-R6: Automated test coverage | ✅ | Added focused unit and functional coverage for the new pathway |

### New files
- `FhirR4Processor.cs` — R4-to-R4 processor with variable support
- `IFhirConverterWithVariables.cs` — interface extending `IFhirConverter` with variable-aware overloads
- `ErrorFilters.cs` — `raise_error` Liquid filter for template-level business rule enforcement
- `DragonPatientMrn.liquid` — sample Dragon MRN transformation template
- `Passthrough.liquid` — minimal template used by shared processor smoke tests
- `metadata.json` — metadata for the `FhirR4` template folder
- `FhirR4ProcessorTests.cs` — unit coverage for processor behavior and edge cases
- `FhirR4ConvertDataFunctionalTests.cs` — end-to-end functional tests
- `ErrorFiltersTests.cs` — unit tests for `raise_error`
- Sample data: `PatientWithIdentifiers.json`, `PatientNoMR.json`, `PatientDuplicateMR.json`, `PatientMultiCodingMR.json`
- Expected output: `DragonPatientMrn.json`

### Modified files
- `DataType.cs` — added `FhirR4`
- `DefaultRootTemplateParentPath.cs` — added `FhirR4`
- `FhirConverterErrorCode.cs` — added `InvalidVariableName = 1314`
- `ConvertProcessorFactory.cs` — added `FhirR4` processor routing
- `Resources.resx` — added variable validation error strings
- `ProcessorTests.cs` — added FhirR4 to shared processor coverage
- `ConvertProcessorFactoryTests.cs` — added FhirR4 factory and interface assertions
- `DefaultTemplateCollectionProviderTests.cs` — scoped assertions to currently packaged template sets
- `TestConstants.cs` — added FhirR4 template directory constant

### Focused FhirR4 coverage
- Added 35 unit tests for:
  - happy path conversion
  - field preservation
  - zero-match and multi-match failures
  - multi-coding single-identifier edge case
  - sparse identifier shapes
  - null/empty variable handling
  - variable name validation
  - missing/empty `dragonSystem`
  - factory and cancellation behavior

- Added 5 functional tests for:
  - successful Dragon identifier generation
  - no-match failure
  - duplicate-match failure
  - factory-based invocation
  - missing `dragonSystem` failure

### Full solution verification
| Suite | Passed | Failed |
|---|---|---|
| Converter Unit Tests | 468 | 0 |
| TemplateManagement Unit Tests | 324 | 0 |
| Converter Functional Tests | 2410 | 0 |

Build verification:
- `dotnet build --warnaserror` → 0 warnings, 0 errors

### CLI tool integration
CLI integration was not added in this PR.

Reason:
- The Phase 1 spec defines library-level requirements only.
- The CLI tool is not called out in Phase 1 requirements.
- The current CLI has no variable-passing mechanism such as a `--var` argument model.
- Routing `FhirR4` through the CLI without a variable input contract would not provide a usable end-to-end experience.

This should be handled as a separate follow-up if CLI support is required.

### Default template packaging / embedded template flows
Default template packaging was not added in this PR.

Reason:
- The spec first introduces default template collection usage in Phase 3 via `templateCollectionReference`.
- FhirR4 templates in this phase depend on runtime variables.
- The current default/embedded template flow has no mechanism to provide those variables.
- Packaging FhirR4 templates into the default template bundle in Phase 1 would add templates that cannot be meaningfully used through that path yet.

This should be addressed alongside API/default-template flow support in a later phase.

### Adding variable overloads to `IFhirConverter`
The base `IFhirConverter` interface was intentionally not changed.

Reason:
- Adding variable overloads there would force existing non-variable processors to implement methods they do not use.
- `IFhirConverterWithVariables : IFhirConverter` keeps the new capability isolated to processors that support it.
- Callers can discover support via interface cast/check, which is the intended design for this phase.

## Notes

This PR is intended to complete the Phase 1 library slice:
- new FHIR R4-to-R4 processor
- variable injection into Liquid templates
- Dragon MRN sample template
- rendering error support
- focused automated validation

It does not attempt to complete later-phase integration surfaces.


src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirR4Processor.cs
```diff
@@ -0,0 +1,185 @@
...
+        private string InternalConvertWithVariables(string data, string rootTemplate, ITemplateProvider templateProvider, IDictionary<string, string> variables, TraceInfo traceInfo = null)
+        {
+            if (string.IsNullOrEmpty(rootTemplate))
+            {
+                throw new RenderException(FhirConverterErrorCode.NullOrEmptyRootTemplate, Resources.NullOrEmptyRootTemplate);
+            }
+
+            if (templateProvider == null)
+            {
+                throw new RenderException(FhirConverterErrorCode.NullTemplateProvider, Resources.NullTemplateProvider);
```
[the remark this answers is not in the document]
dustinburson: If too complex can be skipped for this iteration but adding to the base converter would allow context to work across conversion types.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirR4Processor.cs
```diff
@@ -0,0 +1,185 @@
...
+                }
+
+                foreach (var kvp in variables)
+                {
+                    ValidateVariableName(kvp.Key);
+
+                    if (kvp.Value != null && kvp.Value.Length > MaxVariableValueLength)
+                    {
+                        throw new RenderException(FhirConverterErrorCode.InvalidVariableName, string.Format(Resources.VariableValueTooLong, kvp.Key, MaxVariableValueLength));
+                    }
```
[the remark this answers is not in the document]
dustinburson: Agree with AI comment, more specific error codes are warrented.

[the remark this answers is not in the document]
dustinburson: Another alternative is to make the error code more generic i.e. Invalid Variable and let the specific message determine the error .

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/IFhirConverterWithVariables.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Threading;
+using Microsoft.Health.Fhir.Liquid.Converter.Models;
+
+namespace Microsoft.Health.Fhir.Liquid.Converter.Processors
+{
+    public interface IFhirConverterWithVariables : IFhirConverter
```
dustinburson: Since we will eventually add this to all IFhirConverters I would suggest adding this to the base interface.  For existing classes they can throw not implemented exceptions.

src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/Providers/DefaultTemplateCollectionProviderTests.cs
```diff
@@ -50,7 +50,7 @@ public async Task GivenDefaultTemplateProvider_WhenGetTemplateCollectionAsync_Th
             Assert.Single(templateCollection);
 
             // Verify expected number of templates per data type as per templates packaged from the data/Templates directory.
-            foreach (var defaultRootTemplateParentPath in Enum.GetValues<DefaultRootTemplateParentPath>())
+            foreach (var defaultRootTemplateParentPath in _defaultTemplatesFolderInfo.Keys)
```
dustinburson, line 53: More context on the necessity of this change is needed.  This change looks to be unrelated to the PR.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirR4Processor.cs
```diff
@@ -0,0 +1,185 @@
...
+using Microsoft.Health.MeasurementUtility;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Liquid.Converter.Processors
+{
+    public class FhirR4Processor : JsonProcessor, IFhirConverterWithVariables
+    {
+        private static readonly Regex ValidVariableNameRegex = new Regex(@"^[a-zA-Z_][a-zA-Z0-9_]*$", RegexOptions.Compiled);
+        private static readonly HashSet<string> ReservedVariableNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "msg", "vars" };
```
dustinburson: Similar to the AI comment but it would be good to separate/decouple the vars parsing and validation logic so it isn't embedded only in this processor.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/IFhirConverterWithVariables.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public string Convert(string data, string rootTemplate, ITemplateProvider templateProvider, IDictionary<string, string> variables, TraceInfo traceInfo = null);
+
+        public string Convert(string data, string rootTemplate, ITemplateProvider templateProvider, IDictionary<string, string> variables, CancellationToken cancellationToken, TraceInfo traceInfo = null);
```
dustinburson: This is fine for a first iteration but one downside of just taking the variables as a dictionary is you can't control if they are case sensitive or in sensitive.  Ideally that isn't defined by the caller.  Internally in the processor the reserved variables are case insensitive and you would want to be consistent with that decision.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirR4Processor.cs
```diff
@@ -0,0 +1,185 @@
...
+            if (template == null)
+            {
+                throw new RenderException(FhirConverterErrorCode.TemplateNotFound, string.Format(Resources.TemplateNotFound, rootTemplate));
+            }
+
+            // Build context with msg + vars as sibling keys
+            var dictionary = new Dictionary<string, object> { { DataKey, jsonData } };
+            if (variables != null && variables.Count > 0)
+            {
+                dictionary["vars"] = Hash.FromDictionary(variables.ToDictionary(kv => kv.Key, kv => (object)kv.Value));
```
dustinburson, line 113: Is there a reason for this to go from a dictionary of string/string to string/object and then be wrapped in a Hashset before being assigned to the context.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirR4Processor.cs
dustinburson: Naming is confusing.  Originally thought you were using as reserved values in the provided vars and would error if they were used.  That isn't the case, these are the names used for the context dictionary that is in scope for the template.  Suggest adding TemplateContext like ReservedTemplateContextVariableNames to be distinguish.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirR4Processor.cs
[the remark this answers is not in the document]
Chethan77: Addressed in a26187d. Went with your suggestion — renamed InvalidVariableName → InvalidVariable as a single generic error code. The specific error message (too many variables, value too long, invalid name format, reserved name, duplicate name) carries the detail. This avoids proliferating error codes for every validation scenario while keeping the messages actionable.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirR4Processor.cs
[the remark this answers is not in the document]
Chethan77: Agreed — skipping for this iteration as suggested.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirR4Processor.cs
```diff
@@ -0,0 +1,185 @@
...
+            if (!ValidVariableNameRegex.IsMatch(name))
+            {
+                throw new RenderException(FhirConverterErrorCode.InvalidVariableName, string.Format(Resources.InvalidVariableNameFormat, name));
+            }
+
+            if (ReservedVariableNames.Contains(name))
+            {
+                throw new RenderException(FhirConverterErrorCode.InvalidVariableName, string.Format(Resources.ReservedVariableName, name));
+            }
+        }
```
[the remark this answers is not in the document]
Chethan77: Addressed in a26187d. ValidateVariableName is now internal on a new VariableValidator utility class under Utilities/. Test access is via InternalsVisibleTo in the .csproj. The method is no longer on FhirR4Processor at all.

src/Microsoft.Health.Fhir.Liquid.Converter/Filters/ErrorFilters.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Microsoft.Health.Fhir.Liquid.Converter
+{
+    /// <summary>
+    /// Filters for raising errors from templates
+    /// </summary>
+    public partial class Filters
+    {
+        public static string RaiseError(string message)
+        {
+            throw new RenderException(FhirConverterErrorCode.TemplateRenderingError, message);
```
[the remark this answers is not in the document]
Chethan77: Addressed in a26187d. Null/empty/whitespace messages are now normalized to a resource string (Resources.DefaultTemplateError → "Template raised an error.") rather than a hardcoded literal, consistent with the project's localization pattern. Tests assert the exact normalized message.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirR4Processor.cs
Chethan77: Addressed in a26187d. Validation logic has been extracted to Utilities/VariableValidator.cs — a shared internal static class with:

ValidateVariables(IDictionary<string, string>) — count limit, per-key name validation, per-value size limit, case-insensitive duplicate detection
ValidateVariableName(string) — null/empty, length, regex format, reserved name check
FhirR4Processor now calls VariableValidator.ValidateVariables(variables) — one line. Any future processor that supports variables reuses the same validator.

src/Microsoft.Health.Fhir.TemplateManagement.UnitTests/Providers/DefaultTemplateCollectionProviderTests.cs
Chethan77: Context: Adding DataType.FhirR4 to the enum caused the previous test to break because it iterated all DefaultRootTemplateParentPath enum values and expected each to have a packaged default template collection. FhirR4 is intentionally not packaged as a default template set in Phase 1 — its templates require runtime variables, and the default/embedded template flow has no variable injection mechanism yet.

The fix switches from enum iteration to an explicit dictionary of the 5 currently-packaged template families. This is more correct — it tests what is actually shipped rather than assuming every enum value maps to a packaged template. When FhirR4 templates are added to the default package in a later phase, that work would add the entry here alongside the packaging change.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirR4Processor.cs
Chethan77: > Added an inline comment in a26187d explaining the chain. The three conversions are each forced by a different layer:
> 1. `Dictionary<string, string>` (public API) — Phase 1 spec restricts variables to string values
> 2. `.ToDictionary(... (object)kv.Value)` — `Hash.FromDictionary` requires `IDictionary<string, object>`
> 3. `Hash.FromDictionary()` → `Hash` — DotLiquid's rendering context requires `Hash` for nested property resolution (`{{ vars.dragonSystem }}`)
>
> Keeping the public API as `string,string` is intentional to enforce the Phase 1 constraint. Phase 2 may widen to `string,object` for typed variables.

src/Microsoft.Health.Fhir.Liquid.Converter/Processors/FhirR4Processor.cs
Chethan77: Addressed in a26187d. Renamed to ReservedTemplateContextVariableNames in the new VariableValidator class. The name now clearly communicates that these are the template rendering context's reserved keys ("msg" for input data, "vars" for the variable namespace), not restrictions on user-provided variable names in general.

